PR #2, opened 2022-11-03
Persistence API #2x1z4ae

cerebral-314:


SavageGameDesign:
Task linked: [CU-2x1z4ae Persistence API basics that supports multiple backends, backed by profile](https://app.clickup.com/t/2x1z4ae)

veteran29 (changes requested):
I'm doing another pass on this Pull Request and I'm not happy with it. I think you made the implementation too specific and focused on the player persistence instead of generic persistence feature.

### General design considerations:

Going over features described in the task:

**1.**
> Basic key-value store functionality (save key with data, load data from key)

There's no wrapper functions for this functionality at all, instead we have direct usage of `profileNamespace` or SQL queries via extDB3. Thinking about how the gamemodes in arma usually work I _feel_ like we should generic functions with signatures like:
```sqf
[<table>, <id>] call vgm_fnc_db_get;
[<table>, <id>, <data>] call vgm_fnc_db_set;

[] call vgm_fnc_db_commit; // saves missionProfileNamespace, in case of remote DB might do nothing depending if we want to use transactions or not
[] call vgm_fnc_db_clear; // most likely only available for missionProfileNamespace
```

**2.**
> Single API, multiple backends - we should be able to use profiledb initially, and easily swap to a shared database later

It's somewhat(ish) here but does not feel like it's abstracted enough (due to above point), SQL queries plainly in the functions that use the DB. 

Personally, I feel like we should not implement external database right now. The current extDB3 implementation feels like a technical debt to me already. Storing text (sqf array) data in single column is pretty meh and would be bit annoying to use in let's say external database with player stats etc.

If and when we will want external storage we IMHO should have custom solution, which could be quickly built with something like `arma-rs` which could allow us to use proper models on extension side if we want, has neat features for converting Rust structs to Arma data and opposite and is quite fast to develop in.

**3.**
> Initial implementation using profile or mission profile.

We should use missionProfileNamespace as this is better for people local hosting, bloating player profile is bad as it can cause stutters. There's no reason to use `profileNamespace` over `missionProfileNamespace`.

**4.**
> Supports clearing all data for easy testing / mission resets

We lack generic implementation so does not apply r/n.

**5.**
> Accepts hashmaps, arrays, strings, numbers as a minimum

We lack generic implementation so does not apply r/n.

#### Example:

Let's say we have a custom _playerSkills_ component, to get known skills on the player load we would have function like:
```sqf
vgm_fnc_playerSkills_get = {
    params ["_playerUID"];

    ["player_skills", _playerUID] call vgm_fnc_db_get // return
};
```

For `missionProfileNamespace` the implementation of `db_get` could be simply:
```sqf
vgm_fnc_db_get = {
    params ["_key", "_id", "_data"];

    missionProfileNamespace getVariable [format ["vgm_%1_%2", _key, _id]] // return
};
```

For the database, be it either some document store like Mongo or relational DB like Postgres it will `callExtension`. Simple and should work fine for 99% of use cases.

### General code style comments:

The braces are not consistent, some parts have brace in same line as `then` some on next line. We should prefer same line for more compact code.

Wayyy too many comments over every single thing, it goes to the point where it darkens the code.
Simple and obvious things do not need comments, if something is not clear we should try using descriptive variable names instead, if possible ofc.

Example of the comments which are not that useful:

// we can see it's an empty array
```sqf
// create empty array for playerVars
_vars = [];
```

// quite obvious, what the comment states is litteraly part of the command name
```sqf
// Get player's UID
private _playerUID = getPlayerUID player;
```

// good idea with the variable name, it's descriptive, so no need for a comment that conveys the same thing
```sqf
// Create a new HashMap to return with
private _returnHashMap = createHashMap;
```

// this is simply false comment, it's last line of the file, thus it's a return.
```sqf
// Add all player variables to the return HashMap
_profile;
```
etc.


As SQF has no explicit return statement I find it good practice to mark them with `// return` comments, like this:
```sqf
fnc_returnExample = {
    if (_something) exitWith {
        123 // return
    };

    312 // return
};
```

game/functions/core/server/db/fn_db_get.sqf
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+params ["_key", "_id"];
+
+private _variable = format ["vgm_%1_%2", _key, _id];
+private _value = missionProfileNamespace getVariable [_variable, []];
+
+private _result = createHashMapFromArray _value;
+
+_result // return
```
Spoffy: I've been thinking about this, and I'm wondering if we should have two commands: a raw DB set/get command, and ones that do key and id.

My thinking is we might want a lower level of DB access in the future, but still without having to deal with the missionProfileNamespace directly.

So maybe something like:
```
[_key, _value] call vgm_fnc_db_save;
["player", _key, _value] call vgm_fnc_db_save_type;
```
Or similar? What do you think?

Spoffy: (Also - my naming there is... a bit meh, so if you've better ideas I'm all ears!)

game/functions/core/server/db/fn_db_set.sqf
```diff
@@ -0,0 +1,38 @@
...
+
+        _playerSkills set ["driving", 1];
+
+        ["player", _playerUID, _playerProfile] call vgm_s_fnc_db_set;
+*/
+
+params ["_key", "_id", "_data"];
+
+if !(_data isEqualType createHashMap) exitWith {
+	diag_log format ["VGM: Failure to save %1(%2). Data is not a hashmap: %2", _key, _id, _data];
```
Spoffy: For logging in general, can we use https://github.com/Savage-Game-Design/Paradigm/blob/development/global/functions/logging/fn_log.sqf instead of diag_log? As it'll give us tool integration later :)

I like the check here, but I'm wondering if we need to restrict it to just hashmaps? Profile namespace supports arbitrary variable types, so it feels like we should be safe allowing the user to store whatever they want.

game/functions/core/server/db/fn_db_set.sqf
```diff
@@ -0,0 +1,38 @@
...
+};
+
+if !(_data find "version") then {
+	_data set ["version", getText(missionConfigFile >> "version")];
+};
```
Spoffy: I like what you've gone for here. I'm wondering if we want to do versioning per-entity, or whether it would be better to do versioning on the whole save, and just upgrade the whole save when we push a new version out. It feels like that might be simpler to track and manage going forwards?

Would like your thoughts on it though, as I'm definitely not sure.

Spoffy (changes requested):

game/functions/core/server/db/fn_db_get.sqf
cerebral-314: I think it'd be a good idea... I could include the key type inside the value just to ensure you'll be using the proper schema when/if we switch backends.

game/functions/core/server/db/fn_db_set.sqf
cerebral-314: Alrighty will swap to fn_log.

I'd like it to remain only hash maps because hash maps can hold arbitrary variable types as well. It just seems proper for a key/value database structure.
With the current configuration it'd be very easy to translate this to any database type. It outputs as a array which is very JS friendly(assuming we head in the MongoDB direction with the extension... assuming we do that as well haha).

game/functions/core/server/db/fn_db_set.sqf
cerebral-314: Yeah definitely... my thought process was what if we wanna "upgrade" the player structure via SQL. Would be a real quick query to separate the old from the new.

game/functions/core/server/db/fn_db_get.sqf
Spoffy: +1. I like the idea of including the key type in the hashmap too.

game/functions/core/server/db/fn_db_set.sqf
Spoffy: Cool - that's fair, I don't really have a strong argument for not using hashmaps. I *suspect* we'll end up writing wrappers like: `vgm_fnc_db_save_player` and `vgm_fnc_db_save_all_vehicles`, etc, anyway, so the hashmap vs arbitrary value debate wouldn't make a difference to the usual usage anyway.

game/functions/core/server/db/fn_db_set.sqf
Spoffy: Yeah, the more I think about it, the more saving a version per entity makes sense, if we end up with multiple servers sharing one database. 

I think versioning the whole DB would cause us issues with having to shut down all the servers and upgrade the whole DB of content before being able to upgrade the servers.

Can I propose though that when we split the `db_set` functions into `db_save` and `db_save_type` variants, we only include the `version` tag insertion in the `db_save_type` function? :)

game/functions/core/server/db/fn_db_set.sqf
Spoffy: (To be clear - that's a 👍 for hashmaps)

game/functions/core/server/db/fn_db_get.sqf
Spoffy: We can remove the return comment I think, the lack of semicolon is what we usually use to distinguish that :)

cerebral-314: Filip told me to do this but I will :P

game/functions/core/server/db/fn_db_save.sqf
```diff
@@ -0,0 +1,39 @@
...
+        [_playerUID, _playerProfile] call vgm_s_fnc_db_save;
+*/
+
+params ["_id", "_data"];
+
+if !(_data isEqualType createHashMap) exitWith {
+    ["ERROR", format ["VGM: Failure to save %1. Data is not a hashmap: %2", _id, _data]] call para_g_fnc_log;
+};
+
+if (_data find "key" == -1) exitWith {
```
Spoffy: I'm thinkg this function shouldn't check key - I think the idea for this function is that `_id` is the full key.
E.g, on line 39, we should have:
`format ["vgm_%1", _id]`.

Then db_save_type calls:
```
[format ["%1_%2", _type, _id], _value] call vgm_fnc_db_save
```

If that makes sense? As right now, I think both functions are more or less doing the same thing. My thinking is that this then gives you the option of going "Nah, forget types, this should just be written to the DB as-is". For example, we might want to do:
```
["version", createHashMapFromArray [["number", "3.2"]]] call vgm_fnc_db_save
```
And also
```
["player", getPlayerUID _player, createHashMap] call vgm_fnc_db_save_type
```

Spoffy: Hopefully the examples get across the difference in utility between the two? :)

game/functions/core/server/db/fn_db_save_type.sqf
```diff
@@ -0,0 +1,44 @@
...
+*/
+
+params ["_key", "_id", "_data"];
+
+if !(_data isEqualType createHashMap) exitWith {
+    ["ERROR", format ["VGM: Failure to save %1. Data is not a hashmap: %2", _id, _data]] call para_g_fnc_log;
+};
+
+// don't overwrite existing key
+if !(_data find "key" == -1) then {
```
Spoffy: We can use `"key" in _data` to check if a key exists in a hashmap :)

cerebral-314: I keep forgetting about that one lmao sorry.

game/functions/core/server/db/fn_db_clear.sqf
```diff
@@ -0,0 +1,23 @@
...
+    Returns: nothing
+
+    Example(s):
+        call vgm_s_fnc_db_clear;
+        
+*/
+
+{
+	missionProfileNamespace setVariable [_x, nil];
+} forEach allVariables missionProfileNamespace;
```
veteran29: Missing EOL.

game/functions/core/server/db/fn_db_clear.sqf
```diff
@@ -0,0 +1,23 @@
...
+    Last Update: 2022-11-11
+    Public: No
+
+    Description:
+        Clears the database of all entries.
+
+    Parameter(s):
+        N/A
+
+    Returns: nothing
```
veteran29: If we want to match the style we use in the DLC (and the one added to automatic file headers extension, ref #5 )
```suggestion
    Returns:
        Nothing
```

game/functions/core/server/db/fn_db_save.sqf
```diff
@@ -0,0 +1,36 @@
...
+    Public: No
+
+    Description:
+        Creates or updates a database entry from a key.
+
+    Parameter(s):
+        0: ID   - ID of the entry IE: player_76563326239693 [STRING]
+        1: Data - Data for the entry [HASHMAP]
+
+    Returns: nothing
```
veteran29:
```suggestion
    Returns:
        Nothing
```

game/functions/core/server/db/fn_db_save_type.sqf
```diff
@@ -0,0 +1,50 @@
...
+        Nothing
+
+    Example(s):
+        private _playerUID = getPlayerUID player;
+        private _playerProfile = ["player", _playerUID] call vgm_fnc_db_get;
+        private _playerSkills = _playerProfile get "skills";
+
+        _playerSkills set ["driving", 1];
+
+        ["player", _playerUID, _playerProfile] call vgm_s_fnc_db_save_type;
```
veteran29: I'm not the fan of `save_type` name, does not feel intuitive to me, not sure what would be the better name 😅

game/functions/core/server/db/fn_db_save.sqf
```diff
@@ -0,0 +1,36 @@
...
+        [_vgmUID, _playerProfile] call vgm_s_fnc_db_save;
+*/
+
+params ["_id", "_data"];
+
+if !(_data isEqualType createHashMap) exitWith {
+    ["ERROR", format ["VGM: Failure to save %1. Data is not a hashmap: %2", _id, _data]] call para_g_fnc_log;
+};
+
+missionProfileNamespace setVariable [format ["vgm_%1", _id], _data];
```
veteran29, line 35: I'm wondering if we shouldn't rename this function to something like `save_raw` and do not do any key formating/prefixing in this one.

game/functions/core/server/db/fn_db_save_type.sqf
```diff
@@ -0,0 +1,50 @@
...
+if !("key" in _data) then {
+    _data set ["key", _key];
+};
+
+// add id to data if it doesn't exist
+if !("id" in _data) then {
+    private _vgmID = format ["vgm_%1_%2", _key, _id];
+
+    _data set ["id", _vgmID];
+};
```
veteran29: Should be faster to not use the `if ! in then` (4 commands).
(_insertOnly: Boolean - (Optional, default false) if set to true, set the value only if the key does not exist already_)

```suggestion
private _vgmID = format ["vgm_%1_%2", _key, _id];
// set key and id only if not present in data already
_data set ["key", _key, true];
_data set ["id", _vgmID, true];
```

game/functions/core/server/db/fn_db_save_type.sqf
```diff
@@ -0,0 +1,50 @@
...
+
+_data set ["version", getText(missionConfigFile >> "version")];
+
+private _formattedID = format ["%1_%2", _key, _id];
+[_formattedID, _data] call vgm_s_fnc_db_save;
```
veteran29: **IF** we would change `save` function to `save_raw` as in my suggestion we would have to use the prefixed ID here.
```suggestion
[_vgmID, _data] call vgm_s_fnc_db_save;
```

game/functions/core/server/db/fn_db_get.sqf
```diff
@@ -0,0 +1,43 @@
...
+    case "HASHMAP": {
+        _result = _value;
+    };
+    default
+    {
+        ["ERROR", format ["VGM: Failure to gather data. ID probably overlaps with something that isn't an array or hashmap.", _id, _data]] call para_g_fnc_log;
+    };
+};
+
+_result
```
veteran29: Why do we need to handle arrays here?

Saving API does not support saving anything else than hashmaps. This does not feel needed. If we have garbage data/arrays in save it's broken most likely anyway.

veteran29 (changes requested):
Some minor code formatting suggestions and functionality change ideas.

game/functions/core/server/db/fn_db_save_type.sqf
Spoffy: Yeah, I said that when I suggested the function name to Cerebral. Really not sure what else to call it, but 'save type' sounds like you're saving a type. 

Maybe `typed_save` would be better?

game/functions/core/server/db/fn_db_save.sqf
Spoffy: I'm thinking we leave it for now, and can pull out a `save_raw` later really easily if we find we need it?

game/functions/core/server/db/fn_db_save_type.sqf
Spoffy: Oh, I like this, completely forgot insert only existed as a parameter

game/functions/core/server/db/fn_db_save.sqf
veteran29: I'm not a fan that `save_type` does the same formatting and we do it in two places. I do not insist on name change.

game/functions/core/server/db/fn_db_save_type.sqf
```diff
@@ -0,0 +1,50 @@
...
+};
+
+// add id to data if it doesn't exist
+if !("id" in _data) then {
+    private _vgmID = format ["vgm_%1_%2", _key, _id];
+
+    _data set ["id", _vgmID];
+};
+
+_data set ["version", getText(missionConfigFile >> "version")];
```
veteran29: See #7 
```suggestion
_data set ["version", vgm_version];
```

game/functions/core/server/db/fn_db_get.sqf
cerebral-314: HashMaps can be saved to the missionProfileNamespace. However theyre not persistent. This was just the beginning of serializing/deserializing to an array so the namespace saves it and it actually persists.

game/functions/core/server/db/fn_db_save_type.sqf
cerebral-314: Ah I like this as well... forgive me I'm a noob when it comes to hashmaps. :P

game/functions/core/server/db/fn_db_save_type.sqf
cerebral-314: Ah I was wondering where that disappeared to... Thanks for the correction.

game/functions/core/server/db/fn_db_get.sqf
veteran29: What do you mean they're not persistent, everything that works in `profileNamespace` works in the mission variant too.

Have you actually saved the namespace (`saveMissionProfileNamespace`)?

cerebral-314: Well after testing again it yields different results... baffling. I was calling saveMissionProfileNamespace as well...

I'll go ahead and make it take hashmaps only.

veteran29 (approved):
LGTM, we should get this merged so we can continue with other features.

We can always refine it during the development.

PR #6, opened 2022-11-06
Move user facing strings to stringtable #332dvkq

veteran29:
Moves user-visible strings to stringtable.

Also removes the `version` from the mission config file and moves it to stringtable. The version is a static placeholder for now but it should be set by the build tool.


SavageGameDesign:
Task linked: [CU-332dvkq Move user facing messages to stringtable](https://app.clickup.com/t/332dvkq)

Spoffy (approved):

PR #9, opened 2022-11-28
Add arma run python command and VSCode task #37hg3ca

veteran29:
- title
- add default paradigm path to config example
- fetch system user name automatically
- default config to run with live build of the DLC and Advanced Developer Tools mod (if subscribed)

Might be bit opinionated for my workflow so all feedback is welcome.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -6,7 +6,7 @@
         {
             "label": "Build",
             "type": "shell",
-            "command": "python3 '${workspaceFolder}\\build\\build.py' build --overwrite",
+            "command": "py '${workspaceFolder}\\build\\build.py' build --overwrite",
             "problemMatcher": [],
             "group": {
                 "kind": "build"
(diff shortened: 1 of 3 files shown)


SavageGameDesign:
Task linked: [CU-37hg3ca Add "Launch arma" command](https://app.clickup.com/t/37hg3ca)

build/config.example.py
```diff
@@ -1,12 +1,32 @@
+import os
 from pathlib import Path
 from vgm.pbo import PBO
 
-username = "Spoffy"
+username = os.getlogin()
+arma_profile = r"dev_vgm"
+
+paradigm_path = os.path.join(os.path.dirname(__file__), "../paradigm")
```
Spoffy: Using pathlib would be an easier to read way of representing this, I think:
```suggestion
paradigm_path = Path(__file__).parent /  "paradigm"
```

.vscode/tasks.json
```diff
@@ -11,6 +11,15 @@
             "group": {
                 "kind": "build"
             }
+        },
+        {
+            "label": "Launch arma",
+            "type": "shell",
+            "command": "python3 '${workspaceFolder}\\build\\run.py'",
```
Spoffy: I think just `python` or `py` are the more common aliases on Windows, so it might be worth using one of those for broader compatibility :)

Spoffy (changes requested):

Spoffy (approved):

PR #8, opened 2022-11-13
MikeForce leftovers removal #38g02wh

veteran29:
- title
- Remove harass subsystem
- Remove slingload locality handler (it's now handled by engine)

Bazilion of markers is no more.

PR #10, opened 2022-12-04
Add Shared Hub area limiter #2x1z1m7

veteran29:
- title
- do we want to auto-enable it on mission load?

Depends on: #8

### DIFF
--- a/game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
+++ b/game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
@@ -15,7 +15,7 @@
         Nothing
 
     Example(s):
-        [] call vgm_c_fnc_shared_hub_areaLimiterEnable
+        [] call vgm_c_fnc_sharedHub_areaLimiterEnable
  */
 
 #define STATE_IN_AREA "in_hub"
@@ -23,10 +23,10 @@
 #define TELEPORT_DELAY 5
 #define PLAYER_IN_HUB (player inArea "vgm_shared_hub")
 
-if (!isNull (missionNamespace getVariable ["vgm_shared_hub_areaLimiterScript", scriptNull])) exitWith {};
+if (!isNull (missionNamespace getVariable ["vgm_sharedHub_areaLimiterScript", scriptNull])) exitWith {};
 
-vgm_shared_hub_areaLimiterScript = [] spawn {
-    scopeName "vgm_shared_hub_areaLimiterScript";
+vgm_sharedHub_areaLimiterScript = [] spawn {
+    scopeName "vgm_sharedHub_areaLimiterScript";
 
     private _state = STATE_IN_AREA;
     private _leftHubTime = -1;
@@ -45,7 +45,7 @@ vgm_shared_hub_areaLimiterScript = [] spawn {
                 };
 
                 if (time > (_leftHubTime + TELEPORT_DELAY)) exitWith {
-                    [] call vgm_c_fnc_teleportPlayerToHub;
+                    [] call vgm_c_fnc_sharedHub_teleportPlayerToHub;
                     _state = STATE_IN_AREA;
                     _leftHubTime = -1;
                 };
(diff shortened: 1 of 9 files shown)


SavageGameDesign:
Task linked: [CU-2x1z1m7 Prevent players leaving the shared hub](https://app.clickup.com/t/2x1z1m7)

game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#define STATE_IN_AREA "in_hub"
+#define STATE_OUTSIDE_AREA "outside_hub"
+#define TELEPORT_DELAY 5
+#define PLAYER_IN_HUB (player inArea "vgm_shared_hub")
+
+if (!isNull (missionNamespace getVariable ["vgm_sharedHub_areaLimiterScript", scriptNull])) exitWith {};
+
+vgm_sharedHub_areaLimiterScript = [] spawn {
```
Spoffy, line 28: I think there's a question about whether we leave this here, or whether we put it in our scheduler to keep things as consistent as possible (`scheduler_add_job`). 

We could also potentially do this with triggers, and avoid the infinite loop entirely?

veteran29: I'm not familiar with Paradigm scheduler so I'm not entirely sure when and where it's worth to use it, eg. what problems does it solve? Armas scheduler is not as terrible as people paint it to be if it's not abused and overloaded.

A trigger approach would be fine but the trigger would also need to know when to limit player to its area (could simple disable its simulation I guess?). We don't have any "mission" subsystem that will let us know about that at the moment so it's kinda harder to do.

With scripted approach it will be easier to handle multiple shared hubs if we want to go that route.

game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
```diff
@@ -0,0 +1,56 @@
...
+    scriptName "vgm_sharedHub_areaLimiterScript";
+
+    private _state = STATE_IN_AREA;
+    private _leftHubTime = -1;
+    while {true} do {
+        call {
+            if (_state == STATE_IN_AREA) exitWith {;
```
Spoffy: I think we can get rid of the `call` `exitWith` constructs now, and just use `continue`?

game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
```diff
@@ -0,0 +1,56 @@
...
+                if (!PLAYER_IN_HUB) then {
+                    _state = STATE_OUTSIDE_AREA;
+                    _leftHubTime = time;
+                };
+            };
+
+            if (_state == STATE_OUTSIDE_AREA) exitWith {
+                if (PLAYER_IN_HUB) exitWith {
+                    _state = STATE_IN_AREA;
```
Spoffy, line 44: I think we also need to reset _leftHubTime to -1 here?

veteran29: Good catch, easy to miss such things when doing "statemachine" via loops. Not a biggest fan but I think FSM or custom SM framework would be too big for such use case.

game/functions/systems/shared_hub/client/fn_sharedHub_areaLimiterEnable.sqf
veteran29: That's how I originally wrote it but I did not like that approach in the end as it ends up with every `if` having one line of `continue;` more and `call, exitWith` approch has constant amount of lines "cost". Performance impact is mostly the same.

Spoffy (approved):
Approving! Might be nice to set this up into a generic "Watch this area" system one day using the event system

PR #12, opened 2022-12-14
Add Arma server launch command to run.py #388ef13

veteran29:
- title

### DIFF
--- a/build/arma_server.example.hpp
+++ b/build/arma_server.example.hpp
@@ -1,0 +1,35 @@
+hostname="VGM Development";
+
+loopback = true;
+localClient[] = {"127.0.0.1"};
+
+admins[] = {
+    "76561197993041837", // veteran29
+    "76561198011383725"  // Spoffy
+};
+passwordAdmin = "admin";
+
+voteThreshold = 1;
+voteMissionPlayers = 2;
+allowedVoteCmds[] = {};
+
+
+armaUnitsTimeout = 1;
+allowedFilePatching = 2;
+battlEye = 0;
+persistent = 0;
+verifySignatures = 0;
+
+class Missions {
+    class vgm {
+        template = "vgm\MISSION_NAME";
+        difficulty = "regular";
+        class Params {};
+    };
+};
+
+class AdvancedOptions {
+    LogObjectNotFound = true;
+    SkipDescriptionParsing = false;
+    ignoreMissionLoadErrors = true;
+};
--- a/build/arma_server.example.hpp
+++ b/build/arma_server.example.hpp
@@ -17,7 +17,7 @@ allowedVoteCmds[] = {};
 armaUnitsTimeout = 1;
 allowedFilePatching = 2;
 battlEye = 0;
-persistent = 0;
+persistent = 1;
 verifySignatures = 0;
 
 class Missions {
--- a/build/config.example.py
+++ b/build/config.example.py
@@ -2,10 +2,12 @@ import os
 from pathlib import Path
 from vgm.pbo import PBO
 
+config_dir = Path(__file__).parent
+
 username = os.getlogin()
 arma_profile = r"dev_vgm"
 
-paradigm_path = Path(__file__).parent /  "paradigm"
+paradigm_path = config_dir.parent /  "paradigm"
 mission_output_path = rf"C:\Users\{username}\Documents\Arma 3 - Other Profiles\{arma_profile}\mpmissions\vgm.cam_lao_nam"
 
 output_paths = {
@@ -30,3 +32,5 @@ arma_args = [
     "-debug",
     rf"{mission_output_path}\mission.sqm", # open mission in editor
 ]
+
+arma_server_config_path = Path(__file__).parent /  "arma_server.hpp"
--- a/build/config.example.py
+++ b/build/config.example.py
@@ -33,4 +33,4 @@ arma_args = [
     rf"{mission_output_path}\mission.sqm", # open mission in editor
 ]
 
-arma_server_config_path = Path(__file__).parent /  "arma_server.hpp"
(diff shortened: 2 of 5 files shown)


SavageGameDesign:
Task linked: [CU-388ef13 Add Server launch command to run.py](https://app.clickup.com/t/388ef13)

build/config.example.py
```diff
@@ -2,10 +2,12 @@
 from pathlib import Path
 from vgm.pbo import PBO
 
+config_dir = Path(__file__).parent
+
 username = os.getlogin()
 arma_profile = r"dev_vgm"
 
-paradigm_path = Path(__file__).parent /  "paradigm"
+paradigm_path = config_dir.parent /  "paradigm"
```
Spoffy, line 10: Is this a bug? - we're now doing `Path(__file__).parent.parent`

build/config.example.py
```diff
@@ -30,3 +32,5 @@
     "-debug",
     rf"{mission_output_path}\mission.sqm", # open mission in editor
 ]
+
+arma_server_config_path = Path(__file__).parent /  "arma_server.hpp"
```
Spoffy: Do we want to reuse `config_dir` here since it's now defined on line 5?

build/run.py
```diff
@@ -1,13 +1,63 @@
...
     subprocess.Popen([config.arma_exe_path] + args)
 
+@click.command("server")
+def arma_server_launch():
+    mission_name = Path(config.mission_output_path).name
+    mpmissions_dir = Path(config.arma_exe_path).parent / "mpmissions" / "vgm"
+    mission_output_dir = mpmissions_dir / mission_name
 
-arma_launch()
+    try_create_arma_server_mission_dir(mpmissions_dir, mission_output_dir)
```
Spoffy: Maybe `try_symlink` would be clearer here. instead of `try_create`?

build/run.py
```diff
@@ -1,13 +1,63 @@
...
+def try_create_arma_server_mission_dir(mpmissions_dir, mission_output_dir):
+    try:
+        # ensure subfolder in mpmissions exists
+        if not os.path.exists(mpmissions_dir):
+            os.mkdir(mpmissions_dir)
+        # symlink mission output folder into mpmissions subfolder
+        if not os.path.exists(mission_output_dir):
+            Path(mission_output_dir).symlink_to(config.mission_output_path)
+    except OSError:
+        return print(f"Failed to create symlink from '{mpmissions_dir / mission_output_dir}' to '{config.mission_output_path}'")
```
Spoffy: Can I suggest we re-throw with a generic exception instead of just returning? That way if something goes wrong, we don't go ahead and still start Arma, which might a nuisance.

build/run.py
```diff
@@ -1,13 +1,63 @@
...
 
+@click.command("server")
+def arma_server_launch():
+    mission_name = Path(config.mission_output_path).name
+    mpmissions_dir = Path(config.arma_exe_path).parent / "mpmissions" / "vgm"
+    mission_output_dir = mpmissions_dir / mission_name
 
-arma_launch()
+    try_create_arma_server_mission_dir(mpmissions_dir, mission_output_dir)
+    try_update_arma_server_config(mission_name)
```
Spoffy: Similarly here, maybe `setup_temporary_arma_server_config`? We can also drop the `try` I think, as if this throws an OS error we abort the whole script (it doesn't fail gracefully)

build/run.py
```diff
@@ -1,13 +1,63 @@
...
+
+    args = [
+        "-mod=" + ";".join(config.arma_arg_mods),
+        "-server",
+        f"-config={config.arma_server_config_path.with_suffix('.tmp')}",
```
Spoffy: Maybe we could use a return value from `setup_temporary_arma_server_config`?

Spoffy (changes requested):

build/run.py
```diff
@@ -1,13 +1,63 @@
...
+        # symlink mission output folder into mpmissions subfolder
+        if not os.path.exists(mission_output_dir):
+            Path(mission_output_dir).symlink_to(config.mission_output_path)
+    except OSError:
+        return print(f"Failed to create symlink from '{mpmissions_dir / mission_output_dir}' to '{config.mission_output_path}'")
+
+def try_update_arma_server_config(mission_name):
+    with open(config.arma_server_config_path, "r") as f:
+        server_config = f.read()
+        server_config = server_config.replace("MISSION_NAME", mission_name)
```
Spoffy, line 51: I like this bit, setting it up dynamically with the mission name.

build/arma_server.example.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+voteThreshold = 1;
+voteMissionPlayers = 2;
+allowedVoteCmds[] = {};
+
+
+armaUnitsTimeout = 1;
+allowedFilePatching = 2;
+battlEye = 0;
+persistent = 0;
```
Spoffy: Would persistent=1 be a better option here, to make it easier to test things like JIP issues?

veteran29: Agreed

build/config.example.py
veteran29: I think the old example config had an error, `Path(__file__).parent` throws an error.
```
config_dir = Path(__file__).parent // build dir with build configs
config_dir.parent // repo root
```

build/run.py
veteran29: Agreed

build/config.example.py
Spoffy: Ah, you're right now I look at the path. `parent` would be `build`, so `parent.parent` would be repo root.

Spoffy (approved):

PR #14, opened 2023-01-03
Get profile and save function for players #30zubgc

cerebral-314:


SavageGameDesign:
Task linked: [CU-30zubgc Give players a personal save](https://app.clickup.com/t/30zubgc)

game/functions/core/server/player/fn_player_get_profile.sqf
```diff
@@ -0,0 +1,40 @@
...
+
+private _profile = ["player", getPlayerUID _player] call vgm_s_fnc_db_get;
+
+if (isNil "_profile") then {
+    _profile = ["player", getPlayerUID _player, createHashMap] call vgm_s_fnc_db_typed_save;
+
+    _profile set ["name", name _player];
+    _profile set ["uid", getPlayerUID _player];
+
+    [_profile] call vgm_s_fnc_player_save;
```
Spoffy: I think I'd expect `vgm_s_fnc_player_save` to accept a player as a parameter, and save them. Having to pass in the profile, means whenever want to save the player, we'll probably end up doing something like:

```sqf
private _profile = player getVariable "vgm_player_profile";
if (isNil "_profile") exitWith {};
[_profile] call vgm_s_fnc_player_save
```

I think it should probably be:
```sqf
[_player] call vgm_s_fnc_player_save;
```

And just create a new hashmap, populate it with relevant info, and then save it to the DB. At least for now :)

game/functions/core/server/player/fn_player_save.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+params ["_playerProfile"];
+
+// Ensure that _playerProfile is a hashmap and it has a uid
+if (typeName _playerProfile != "HASHMAP" || isNil _playerProfile get "uid") exitWith {
+    ["ERROR", format ["Invalid player profile: %1", _playerProfile]] call para_g_fnc_log;
+    false
+};
+
+["player", getPlayerUID player, _playerProfile] call vgm_s_fnc_db_typed_save;
```
Spoffy: This won't work as it's being run on the server, so `player` won't be the player to save :). But see my comment above first before mamiing any changes here anyway!

game/functions/functions_server.hpp
```diff
@@ -25,4 +25,12 @@ class vgm_s
         class db_save {};
         class db_typed_save {};
     };
+
+    class player
+    {
+        VGM_SERVER_PATH(\core\server\player);
+
+        class player_get_profile {};
```
Spoffy: I think using the word 'profile' is going to get confusing here, since the term `player profile` already refers to their local profile file.

Maybe just `fetch_save` would be good. I prefer `fetch` over `get` as I think `fetch` implies doing more work than just returning a variable, such as accessing the DB!
// Think about appropriate naming/functionality for a player save system

Spoffy (changes requested):

game/functions/core/server/db/fn_db_get.sqf
```diff
@@ -22,5 +22,6 @@
 params ["_key", "_id"];
 
 private _variable = format ["vgm_%1_%2", _key, _id];
+private _result = missionProfileNamespace getVariable [_variable, createHashMap];
 
-missionProfileNamespace getVariable [_variable, createHashMap]
+_result
```
Spoffy: Let's add the deep copy here (`+_result`)

game/functions/core/server/player/fn_player_fetch.sqf
```diff
@@ -1,8 +1,8 @@
 /*
-    File: fn_player_get_profile.sqf
+    File: fn_player_get_fetch.sqf
```
Spoffy: The file name here isn't consistent with the actual file name :)

game/functions/core/server/player/fn_player_fetch.sqf
```diff
@@ -28,13 +28,14 @@ if (!isPlayer _player) exitWith {
...
-if (isNil "_profile") then {
+// Check if hashmap is empty
+if ("uid" in _profile == false) then {
     _profile = ["player", getPlayerUID _player, createHashMap] call vgm_s_fnc_db_typed_save;
 
     _profile set ["name", name _player];
     _profile set ["uid", getPlayerUID _player];
 
-    [_profile] call vgm_s_fnc_player_save;
+    [_player] call vgm_s_fnc_player_save;
```
Spoffy: I don't think we should be doing two saves here.
We do a save - then set some properties, then do a save.
It feels like we should just invoke `[_player] call vgm_s_fnc_player_save`, which creates the save if it doesn't exist and returns it.

My reasoning is that we have double the number of saves, and we also have two places that are writing to the "save format" for players, which makes it easy to introduce inconsistencies.

game/functions/core/server/player/fn_player_save.sqf
```diff
@@ -2,35 +2,33 @@
...
 
-// Ensure that _playerProfile is a hashmap and it has a uid
-if (typeName _playerProfile != "HASHMAP" || isNil _playerProfile get "uid") exitWith {
-    ["ERROR", format ["Invalid player profile: %1", _playerProfile]] call para_g_fnc_log;
-    false
-};
+if (!isPlayer _player) exitWith { ["ERROR", format ["Object is not a player: %1", _player]] call para_g_fnc_log; };
 
-["player", getPlayerUID player, _playerProfile] call vgm_s_fnc_db_typed_save;
+private _playerProfile = [_player] call vgm_s_fnc_player_fetch;
```
Spoffy: I don't think we need to fetch the profile from storage here. I think the safest thing to do is regenerate the save from scratch. That way we're not worrying about old data, or the existence/non-existence of the save.

What do you think?

Spoffy (changes requested):

game/functions/core/server/db/fn_db_get.sqf
cerebral-314: Will do!

game/functions/core/server/player/fn_player_fetch.sqf
cerebral-314: Thanks for the spot

game/functions/core/server/player/fn_player_fetch.sqf
cerebral-314: Thanks for the spot on this too... not entirely sure what my thought process was.

game/functions/core/server/player/fn_player_save.sqf
cerebral-314: To be honest this save function is really awkward... The reason I had this function originally take a profile as a parameter was due to this. It'd allow the programmer to fetch the profile do their changes and then save it to the database.

Spoffy: My main thinking here is by having all the save code in one place, it avoids the code that modifies the profile object being spread throughout the codebase. It keeps things isolated, so we don't get ourselves into bug or refactoring hell later when one bit of the codebase touches a bit of the profile it shouldn't! 

In my mind, this function collects all the data associated with the player from the gamemode, and then stores it - which keeps all that modification to one file and gives us some abstraction. 

Which aspect of the save function is making you feel awkward? If we can get to the bottom of that, it might help improve the save system :)

cerebral-314: When we make deep copies of the hash maps we're no longer keeping all the save code in one place. That's what made me wanna trash the save function to begin with since hashmaps "save" when using the set function.

Spoffy: > When we make deep copies of the hash maps we're no longer keeping all the save code in one place

Can you explain this one a little more? 🙂

cerebral-314: Currently the main database hash-map is stored on the missionProfileNamespace. Hash maps are essentially pointers to their objects. 

At the moment if we make a change to the database it all edits 1 place in memory. Which is what I think you want... however if you have me deep copy the hash-map every time I'm allocating another portion in memory that as you say we will have to maintain. 

Sorry for the run on sentence; it's kind of hard to explain. I can explain this with code if need be. A demonstration would be real easy to whip up.

Spoffy: > another portion in memory that as you say we will have to maintain.
Ah, I think this might be where the misunderstanding is happening. I'm proposing we deep copy in db_get, then return the deep copy. We use the returned deep copy in a read-only way to populate the right systems, then throw that copy away.

I'm not proposing keeping that copy around long-term. My reason for deep copying is it protects us from accidental changes to the original copy, *and* gives us the same behaviour we'd see when working with a database (a copy of the data, where modifying it doesn't affect source data)

Spoffy: In essence, we do:

1. db_get (retrieve raw data from db)
2. player_fetch (load player data into relevant parts of the gamemode). 

Once the data has been loaded into the various different bits of the gamemode, we de-allocate/forget about the data that was returned from the DB.

In other words, we effectively load the content from the DB, "de-serialize" it, and let player init continue?

cerebral-314: Okay I think I understand what you want... I'll whip it up real quick

game/functions/core/server/player/fn_player_save.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+if (!isPlayer _player) exitWith { ["ERROR", format ["Object is not a player: %1", _player]] call para_g_fnc_log; };
+
+private _profile = createHashMap;
+
+// Get the player's information from mission
+_profile set ["name", name _player];
+_profile set ["uid", getPlayerUID _player];
+
+private _savedData = ["player", getPlayerUID _player, _player] call vgm_s_fnc_db_typed_save;
```
veteran29: Player object can't be saved to DB, it should be the HashMap.
```suggestion
private _savedData = ["player", getPlayerUID _player, _profile] call vgm_s_fnc_db_typed_save;
```

game/functions/core/server/player/fn_player_save.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+params ["_player"];
+
+if (!isPlayer _player) exitWith { ["ERROR", format ["Object is not a player: %1", _player]] call para_g_fnc_log; };
+
+private _profile = createHashMap;
+
+// Get the player's information from mission
+_profile set ["name", name _player];
+_profile set ["uid", getPlayerUID _player];
```
veteran29, line 33: I don't get this example. This function will always reset the "profile" to new hash, `private _profile = createHashMap;`

game/functions/core/server/player/fn_player_save.sqf
```diff
@@ -0,0 +1,41 @@
...
+    Author: Cerebral
+    Date: 2023-01-03
+    Last Update: 2023-01-08
+    Public: No
+
+    Description:
+        This function will force a save of the player's information to the database.
+
+        This shouldn't be required since the information is a reference to
+            the missionNamespace object, but it's here just in case.
```
veteran29: I don't get the meaning of this. If this is not required why even have it. Code is there to be used or not.

veteran29 (changes requested):

game/functions/core/server/player/fn_player_save.sqf
Spoffy: This was my suggestion - we'll add extra lines in here for data when we need to persist things, whether they're stored on the player object, or in other systems, etc. 

So it'll create a new save, add the relevant data, then save it.

game/functions/core/server/player/fn_player_fetch.sqf
Spoffy: This still looks inconsistent? :D

Spoffy (changes requested):
Looking 99% of the way there to me now :)

game/functions/core/server/player/fn_player_save.sqf
cerebral-314: With the change above this function will be required. I need to update that comment

game/functions/core/server/player/fn_player_save.sqf
Spoffy: We'll use the provider system (scripted event handler wrapper) for this once i've implemented it

game/functions/core/server/player/fn_player_fetch.sqf
```diff
@@ -0,0 +1,31 @@
+/*
+    File: fn_player_get_fetch.sqf
```
Spoffy: This is the last thing to fix for me :D

Spoffy (changes requested):

game/functions/core/server/player/fn_player_fetch.sqf
cerebral-314: Nice catch thank you :P

Spoffy (approved):

veteran29 (approved):
LGTM

PR #15, opened 2023-01-29
Skill system framework #3918y7b

veteran29:
### Acceptance Criteria
- [x] A skill can be defined declaratively in a config or SQF file
- [x] A skill can be unlocked by purchasing one or more other skills (using AND or OR boolean logic) - Callback functions for skills unlocking
- [x] A skill can be manually unlocked by a script. (e.g, when a player completes a certain action or visits a certain location). This is persistent.
- [x] An entire skill tree can be unlocked by one or more skills (e.g, specialization and expertise trees)
- [x] A skill tree has a parent tree, which will be used in the UI to give it a hierarchy.
- [x] A skill tree has a name, and a description, which will be used in the UI.
- [x] A skill can be purchased by spending a skill point, giving the player its benefit.
- [x] A passive skill is made active as soon as it is purchased, and persists across missions and respawn.
- [x] An active skill is triggerable by the player. 
- [x] An active skill has a cooldown period after activation in which it cannot be used.
- [x] The cooldown period affects each type of skill individually.
    - [x] E.g - Activating the primary skill doesn't start the ultimate skill cooldown.
    - [x] E.g - Activating the primary skill then swapping it out, means the new skill is still on cooldown.
- [x] One primary skill may be usable at a time. 
- [x] One ultimate skill may be usable at a time.
- ~~The ultimate skill must be lockable via script to prevent it being changed (e.g, when on a mission)~~
    - Should be handled in UI, IMO - @veteran29
- [x] Skills may have a criteria, which if not met means they will be invisible in the UI. 
- [x] Skills have a description, which will be shown in the UI.
- [x] Skills have a short description, which will be shown in the UI.
- [x] Skills have a name, which is shown in the UI.
- [x] Skills have an icon, which is shown in the UI.
- [x] Skills must be associated with a skill tree


SavageGameDesign:
Task linked: [CU-3918y7b Skills framework](https://app.clickup.com/t/3918y7b)

game/configs/mission/cfg_skills.hpp
```diff
@@ -0,0 +1,120 @@
...
+    description = "SKILL DESCRIPTION";
+    icon = "\vn\ui_f_vietnam\ui\wheelmenu\img\ui_icon_a_ca.paa";
+
+    isActive = 0; // 1 - primary skill, 2 - ultimate skill
+    applyOnRespawn = 0;
+
+    conditionUnlock = "true";
+    conditionShow = "true";
+    codeApply = "systemChat 'skill applied'";
+    codeActivate = "systemChat 'skill activated'";
```
Spoffy: I think we'll need a hook for deactivating too, in case someone changes their skills

Spoffy: We'll also need a cooldown property, for the delay between activations

game/functions/functions_client.hpp
```diff
@@ -3,6 +3,41 @@
...
+        class skills_canLearn {};
+        class skills_canSee {};
+        class skills_getByPath {};
+        class skills_isKnown {};
+        class skills_parseTreeCfg {};
+        class skills_preInit
+        {
+            preInit = 1;
+        };
+        class skills_treesHashToPathsHash;
```
Spoffy, line 55: I like your functions names, feels like I can intuit what these do easily enough.

game/functions/functions_client.hpp
```diff
@@ -28,23 +63,48 @@ class vgm_c
...
+        class skills_active_isSlotOnCooldown {};
+        class skills_active_assignSkillToSlot {};
+        class skills_active_openAssignMenu {};
+        class skills_active_openSkillWheel {};
+        class skills_active_skillWheelActivate {};
+
+        class skills_active_preInit
         {
             preInit = 1;
         };
```
Spoffy: Can we call this preinit from the main skills one? It feels like active skills shouldn't be initialising separately from the rest of the skills system?

I think it makes it confusing for someone to deduce how the init process works. If they're reading the main init file to understand the system, they need to:
1.  Realise there's two init files
2.  Realise they're both configured in the functions config
3.  Understand that the load order is defined by the order in this file, which I don't think is common knowledge.

game/functions/functions_server.hpp
```diff
@@ -25,4 +25,29 @@ class vgm_s
...
+        class skills_teachSkill {};
+    };
+
+    class skills_network
+    {
+        VGM_SERVER_PATH(\systems\skills\server\network);
+
+        class skills_handle_skillsDataRequest {};
+        class skills_handle_skillLearnRequest {};
+    };
```
Spoffy: Nicely organised, I like your folder structure for the skills system.

game/functions/systems/skills/server/fn_skills_teachSkill.sqf
```diff
@@ -0,0 +1,30 @@
...
+    Date: 2023-01-27
+    Last Update: 2023-02-02
+    Public: Yes
+
+    Description:
+        Teach the skill to the player.
+
+    Parameter(s):
+        _player - Player to give the skill point to [OBJECT]
+        _skillPath - Path to the skill being teched [ARRAY]
```
Spoffy: I think it should be `taught` here instead of `teached` :)

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
...
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        _player call vgm_s_fnc_skills_dbGet
+ */
+
+params ["_player"];
+
+private _hash = _player getVariable "vgm_g_skillsData";
```
Spoffy: By _hash, do you mean a skills hash table? As this confused me for a minute, I thought we were passing around some kind of `hashValue _skills` or similar :)

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
...
+if (isNil "_hash") then {
+    _hash = ["player_skills", getPlayerUID _player] call vgm_s_fnc_db_get;
+    _player setVariable ["vgm_g_skillsData", _hash];
+};
```
Spoffy: I think we should load player skills from the DB separately, rather than load it here.

My thinking is:
* If we move to an actual DB in the future, this would potentially be a blocking function (or we'd need to rewrite the db layer to have async options). Which would mean this whole function has a fast path, and a slow path, and we wouldn't know which we'd hit when calling it.
* It pulls from a cache, which I wouldn't have expected based on the name - I'd have expected a db fetch each time.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+_hash getOrDefault ["skillPoints" , 0, true];
+_hash getOrDefault ["skillPaths" , [], true];
```
Spoffy: What do you think to using `_hash set ["skillPoints", 0, true]`, to be a bit a little clearer about what this operation is doing, since we're not using the value it returns?

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
...
+    Example(s):
+        _player call vgm_s_fnc_skills_dbGet
+ */
+
+params ["_player"];
+
+private _hash = _player getVariable "vgm_g_skillsData";
+if (isNil "_hash") then {
+    _hash = ["player_skills", getPlayerUID _player] call vgm_s_fnc_db_get;
+    _player setVariable ["vgm_g_skillsData", _hash];
```
Spoffy: Will there be a runtime error if there's no skills data in the database? Specifically when we hit the `getOrDefault` lines below

game/functions/systems/skills/server/fn_skills_dbSave.sqf
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+params ["_player"];
+
+private _uid = getPlayerUID _player;
+private _hash = _player call vgm_s_fnc_skills_dbGet;
+if (isNil "_hash") exitWith {
+    ["ERROR", format ["VGM: No data for player %1 (%2)", name _player, _uid]] call para_g_fnc_log;
+    false // return
+};
```
Spoffy: Where is the skills data structure initially set on new players? As I can't see dbGet or dbSet handling it, and I can't find the init code elsewhere, but I very well could just be missing it.

game/functions/systems/skills/server/fn_skills_dbSave.sqf
```diff
@@ -0,0 +1,31 @@
...
+
+["player_skills", _uid, _hash] call vgm_s_fnc_db_typed_save;
+saveMissionProfileNamespace // return
```
Spoffy: I think this shouldn't a concern of the skills system, feels like this should be transparently handled by the db system, as we shouldn't need to care about the backend saving implementation.

game/functions/systems/skills/server/fn_skills_dbSave.sqf
```diff
@@ -0,0 +1,31 @@
...
+        Data was saved [BOOL]
+
+    Example(s):
+        _player call vgm_s_fnc_skills_dbSave
+ */
+
+params ["_player"];
+
+private _uid = getPlayerUID _player;
+private _hash = _player call vgm_s_fnc_skills_dbGet;
```
Spoffy: Similar comment to above about naming, I wonder if we should use `_skillsHashMap` or `_playerSkills` or similar?

game/functions/systems/skills/server/fn_skills_teachSkill.sqf
```diff
@@ -0,0 +1,30 @@
+/*
+    File: fn_skills_teachSkill.sqf
```
Spoffy, line 2: I can't find any calls to this function - is it just a utility function for debugging?

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
```diff
@@ -0,0 +1,48 @@
...
+player setVariable ["vgm_g_skillsData", _skillsData];
+
+{
+    private _skill = _x call vgm_g_fnc_skills_getByPath;
+    if (isNil "_skill") then {
+        ["ERROR", format ["VGM: Skill does not exist '%1'", _x]] call para_g_fnc_log;
+        continue;
+    };
+
+    if (_skill getOrDefault ["_known", false]) then {
```
Spoffy: Why do we use `_` at the start of this hashmap property here?

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
```diff
@@ -0,0 +1,48 @@
...
+        ["DEBUG", format ["VGM: Skill known previously '%1'", _x]] call para_g_fnc_log;
+        continue;
+    };
+
+    _skill set ["_known", true];
```
Spoffy: When the player receives skill data, I don't think we roll back their old skills data? So for example, if we un-taught a skill on the server, it would always be set as known on the client.

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
```diff
@@ -0,0 +1,48 @@
...
+
+    ["DEBUG", format ["VGM: New skill '%1'", _x]] call para_g_fnc_log;
+
+    ["vgm_skills_learnt", [_x, _skill]] call para_g_fnc_event_trigger;
```
Spoffy, line 65: If this is called whenever we learn a skill, for every skill, I think that means the `apply` code for each skill would need to be idempotent? 

For example, it couldn't be `player setCustomAimCoef (getCustomAimCoef player + 1)`, every time a player learned a different skill it'd give them the bonus for that skill again.

I think this poses some interesting challenges, such as multiple skills which modify the same thing on the player, but not unsolvable ones.

There's three options I think we could take here:
* Only apply newly learned skills
* Remove all skills, and re-apply them all.
* Add documentation to the skills config that `apply` needs to be idempotent, and explain the consequences.

Given idempotency can be quite a nice property (it'll reduce bugs most likely, although increasing complexity), I'd be keen to hear what you think.

game/functions/systems/skills/server/network/fn_skills_handle_skillLearnRequest.sqf
```diff
@@ -0,0 +1,44 @@
...
+        No description added yet.
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
+ */
+
+params ["_player", "_skillPath"];
```
Spoffy, line 18: Do we need to do `owner _player isEqualTo remoteExecutedOwner` for security here?

game/functions/systems/skills/global/fn_skills_preInit.sqf
```diff
@@ -0,0 +1,19 @@
...
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Nothing
+ */
+
+vgm_skills_treesHash = [missionConfigFile >> "vgm_skillTrees"] call vgm_g_fnc_skills_parseTreeCfg;
+vgm_skills_pathsHash = [vgm_skills_treesHash] call vgm_g_fnc_skills_treesHashToPathsHash;
```
Spoffy: I like that you've cached these in hashmaps, great idea. Wonder if we should add "map" to the end of "Hash" to make it clear we've got a map?

game/configs/mission/cfg_skills.hpp
```diff
@@ -0,0 +1,120 @@
+
+
+class vgm_skillTemplate {
+    displayName = "SKILL NAME";
+    description = "SKILL DESCRIPTION";
+    icon = "\vn\ui_f_vietnam\ui\wheelmenu\img\ui_icon_a_ca.paa";
+
+    isActive = 0; // 1 - primary skill, 2 - ultimate skill
```
Spoffy: Can we use `skillType` or similar for this? Since this is prefixed with `is`, I think I'd expect it to be a boolean value

game/functions/systems/skills/global/fn_skills_parseTreeCfg.sqf
```diff
@@ -0,0 +1,82 @@
...
+                ["displayName", getText (_x >> "displayName")],
+                ["description", getText (_x >> "description")],
+                ["icon", getText (_x >> "icon")],
+                ["isActive", getNumber (_x >> "isActive") > 0],
+                ["isUltimate", getNumber (_x >> "isActive") > 1],
+                ["applyOnRespawn", getNumber (_x >> "applyOnRespawn") > 0],
+                ["conditionUnlock", compileFinal getText (_x >> "conditionUnlock")],
+                ["conditionShow", compileFinal getText (_x >> "conditionShow")],
+                ["codeApply", compileFinal getText (_x >> "codeApply")],
+                ["codeActivate", compileFinal getText (_x >> "codeActivate")]
```
Spoffy: Does compileFinal work in hashmaps? This isn't feedback, I'm just curious :)

game/functions/systems/skills/client/fn_skills_getSkillTreeFromSkill.sqf
```diff
@@ -0,0 +1,32 @@
...
+        _skill call vgm_c_fnc_skills_getSkillTreeFromSkill
+ */
+
+params ["_skill"];
+
+private _path = +(_skill get "path");
+
+private _item = vgm_skills_treesHash get (_path deleteAt 0);
+while {true} do {
+    if (count _path == 1) exitWith {
```
Spoffy: Can we do `<= 1` here? 

On the happy path where everything works, it'll behave the same.

On the unhappy path (where path ends up as 0, if we *somehow* have bad data in skill), it prevents an infinite loop.

Not exactly a likely issue, but it's a little bit more defensive :)

Can add a comment why it's `<=` for clarity too.

game/functions/systems/skills/client/fn_skills_getSkillPoints.sqf
```diff
@@ -0,0 +1,21 @@
+/*
+    File: fn_skills_getSkillPoints.sqf
```
Spoffy, line 2: To be 100% clear, is this the skill points the player has left to spend, and not their total?

game/functions/systems/skills/client/active/fn_skills_active_openSkillWheel.sqf
```diff
@@ -0,0 +1,44 @@
...
+            "vgm_c_fnc_skills_active_skillWheelActivate" // function
+        ]
+    ];
+} forEach [
+    "ultimate",
+    "ability1"
+    // ,"ability2"
+];
+
+[_iconsArray] call vn_fnc_wm_init;
```
Spoffy, line 44: We've got a wrapper around the native wheel menu here, that might be easier to use, and has some convenience functions (such as loading a wheel menu interface from config)
paradigm\client\functions\ui\wheel_menu\fn_wheel_menu_open.sqf

game/functions/systems/skills/client/active/fn_skills_active_preInit.sqf
```diff
@@ -0,0 +1,54 @@
...
+// (grave/tilde `/~ by default)
+addUserActionEventHandler ["selectAll", "Activate", {
+    [] spawn {
+        private _timeout = time + 1;
+        waitUntil {commandingMenu != "" || time > _timeout};
+        showCommandingMenu "";
+    };
+
+    [] call vgm_c_fnc_skills_active_openSkillWheel;
+}];
```
Spoffy: Short term, all good with this.

Long term, we should integrate the keybindings menu from MF, or use Arma's builtin one. I'll open a dev ticket for it.

game/functions/systems/skills/client/active/fn_skills_active_preInit.sqf
```diff
@@ -0,0 +1,54 @@
...
+    ["ultimate", call _fnc_createSlot]
+];
+
+["vgm_skills_learnt", {
+    _this#0 params ["_path", "_skill"];
+    if (!(_skill get "isActive")) exitWith {};
+
+    vgm_c_skills_active_list set [_path, _skill];
+
+}] call para_g_fnc_event_subscribeLocal;
```
Spoffy: We'll need to handle unlearning here too.

game/functions/systems/skills/client/active/fn_skills_active_isSlotOnCooldown.sqf
```diff
@@ -0,0 +1,29 @@
...
+
+params [
+    ["_slot", createHashMap, ["", createHashMap]]
+];
+
+if (_slot isEqualType "") then {
+    _slot = vgm_c_skills_active_slots getOrDefault [_slot, createHashMap];
+};
+
+time < (_slot getOrDefault ["cooldown", 1e10]) // return
```
Spoffy: I'd have expected cooldown to be "how long this takes to cooldown" from the name, such as 60 seconds, or 20 seconds remaining.

Maybe we could use "nextAvailableTime" or "cooldownEndTime" to be clearer :)

game/functions/systems/skills/client/active/fn_skills_active_skillWheelActivate.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+(_this#1) params ["_slot", "_skill"];
+
+if (_slot call vgm_c_fnc_skills_active_isSlotOnCooldown) exitWith {
+    ["WARNING", format ["VGM: Skill on cooldown`", _skill get "path"]] call para_g_fnc_log;
+    hint "Skill on cooldown!";
+};
+
+player call (_skill get "codeActivate");
+private _cooldown = time + ([30, 60] select (_skill get "isUltimate"));
```
Spoffy: Same as the config comment, it'd be good to have this as a per-skill setting :)

game/functions/systems/skills/global/fn_skills_parseTreeCfg.sqf
```diff
@@ -0,0 +1,82 @@
...
+    {
+        private _cfgTier = _cfgSkills >> _x;
+        if (isNull _cfgTier) exitWith {};
+
+        private _tier = _forEachIndex;
+        private _skills = "true" configClasses _cfgTier apply {
+            createHashMapFromArray [
+                ["path", _path + [configName _x]],
+                ["tier", _tier],
+                ["cost", 1],
```
Spoffy: Good thinking, can we pull this out into the config too? :)

game/functions/systems/skills/server/fn_skills_addSkillPoint.sqf
```diff
@@ -0,0 +1,33 @@
...
+["INFO", format ["VGM: Adding skill point to %1 (%2)", name _player, getPlayerUID _player]] call para_g_fnc_log;
+
+private _skillsData = _player call vgm_s_fnc_skills_dbGet;
+private _skillPoints = _skillsData get "skillPoints";
+
+_skillsData set ["skillPoints", _skillPoints + 1];
+
+if (_broadcast) then {
+    [_skillsData] remoteExecCall ["vgm_c_fnc_skills_receiveSkillsData", _player];
+};
```
Spoffy: Wonder if we could/should add a skill point without re-broadcasting all the skills data? As I think this will trigger them all to be re-learned, etc.

game/functions/systems/skills/global/fn_skills_canLearn.sqf
```diff
@@ -0,0 +1,61 @@
...
+    // if first tier check if has previous tree
+    if (_tier < 1) exitWith {
+        private _path = _skill get "path";
+        // no previous tree, always unlocked
+        if (count _path < 3) exitWith {true};
+
+        _path = _path select [0, count _path - 2];
+        private _skillTree = _path call vgm_g_fnc_skills_getByPath;
+        private _tiersCount = count (_skillTree get "skills");
+        diag_log [_skillTree get "displayName", count (_skillTree get "skills")];
```
Spoffy: Rogue diag_log. Maybe we put it as a debug log, or cut it? :)

game/functions/systems/skills/global/fn_skills_canLearn.sqf
```diff
@@ -0,0 +1,61 @@
...
+&& {
+
+    private _tier = _skill get "tier";
+    // if first tier check if has previous tree
+    if (_tier < 1) exitWith {
+        private _path = _skill get "path";
+        // no previous tree, always unlocked
+        if (count _path < 3) exitWith {true};
+
+        _path = _path select [0, count _path - 2];
```
Spoffy, line 48: Could we put a comment in here, explaining we're selecting the skill's tree's parent tree? :)

game/functions/systems/skills/global/fn_skills_canLearn.sqf
```diff
@@ -0,0 +1,61 @@
...
+        private _skillTree = _path call vgm_g_fnc_skills_getByPath;
```
Spoffy: Same here, maybe we rename this to `parentSkillTree` or similar for clarity :)

game/functions/systems/skills/server/fn_skills_addSkillPoint.sqf
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+params ["_player", ["_broadcast", false]];
+
+["INFO", format ["VGM: Adding skill point to %1 (%2)", name _player, getPlayerUID _player]] call para_g_fnc_log;
+
+private _skillsData = _player call vgm_s_fnc_skills_dbGet;
+private _skillPoints = _skillsData get "skillPoints";
+
+_skillsData set ["skillPoints", _skillPoints + 1];
```
Spoffy: Is there a missing skills_dbSave here?

game/functions/systems/skills/server/network/fn_skills_handle_skillLearnRequest.sqf
```diff
@@ -0,0 +1,44 @@
...
+
+["SUCCESS", format ["VGM: Handling learn request for %1 (%2), %3", name _player, getPlayeRUID _player, _skillPath]] call para_g_fnc_log;
+
+// add skill path to known skills
+private _skillsData = _player call vgm_s_fnc_skills_dbGet;
+_skillsData get "skillPaths" pushBackUnique _skillPath;
+
+// decrease amount of skill points
+private _skillPoints = _skillsData get "skillPoints";
+_skillsData set ["skillPoints", _skillPoints - 1];
```
Spoffy: this is currently -1, I think we should be grabbing the skill's cost?

game/functions/systems/skills/server/fn_skills_addSkillPoint.sqf
```diff
@@ -0,0 +1,33 @@
...
+        _broadcast - Should the skill data be sent to the player [BOOL, defaults to false]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        [player, true] call vgm_s_fnc_skills_addSkillPoint
+ */
+
+params ["_player", ["_broadcast", false]];
```
Spoffy: What's the situation where not sync'ing this with the client would help us?  I'm a little concerned the default settings here might lead to us accidentally de-syncing the skills from the client.

game/functions/systems/skills/server/network/fn_skills_handle_skillLearnRequest.sqf
```diff
@@ -0,0 +1,44 @@
...
+
+if (!_canLearn) exitWith {
+    ["WARNING", format ["VGM: Discarding learn request for %1 (%2), %3", name _player, getPlayeRUID _player, _skillPath]] call para_g_fnc_log;
+
+    // send updated skills data to the client
+    [_canLearn, nil, _skillPath] remoteExecCall ["vgm_c_fnc_skills_receiveSkillLearn", _player];
+};
+
+
+["SUCCESS", format ["VGM: Handling learn request for %1 (%2), %3", name _player, getPlayeRUID _player, _skillPath]] call para_g_fnc_log;
```
Spoffy: We probably want to use "INFO" or "DEBUG", as I don't think "SUCCESS" is one of the standard (undocumented) log levels. I'll push to paradigm later with the header updated with a list of log levels.

game/functions/systems/skills/server/fn_skills_teachSkill.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+params ["_player", "_skillPath"];
+
+// add skill path to known skills
+private _skillsData = _player call vgm_s_fnc_skills_dbGet;
+_skillsData get "skillPaths" pushBackUnique _skillPath;
+
+[_player, _skillsData] call vgm_s_fnc_skills_dbSave;
+
+[_skillsData] remoteExecCall ["vgm_c_fnc_skills_receiveSkillsData", _player];
```
Spoffy: Feels like this duplicates handle_skillLearnRequest a lot. Maybe we should have handle_skillLearnRequest call this? (Wondering if that was your original plan, since this is unused currently)

Spoffy (changes requested):
Phew, big PR! Left a bunch of feedback. I don't think there's major or system changing, just lots of little tweaks and a few bits of added functionality.

The biggest thing I think we need to add is unlearning skills. I forgot to spec it out in the original ticket I think ☹️

game/functions/systems/skills/server/fn_skills_addSkillPoint.sqf
veteran29: The idea was that it can be executed multiple times before broadcasting but it does not make sense, I've added `_amount` param instead if you want to add multiple skill points.

game/functions/systems/skills/server/fn_skills_teachSkill.sqf
veteran29: Refactored it to use `teachSkill` in `skillLearnRequest`.

game/functions/systems/skills/server/network/fn_skills_handle_skillLearnRequest.sqf
veteran29: Skill cost wasn't fully implemented, fixed.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
veteran29: It's a hashMap

game/functions/systems/skills/server/fn_skills_teachSkill.sqf
veteran29: >A skill can be manually unlocked by a script. (e.g, when a player completes a certain action or visits a certain location). This is persistent.

API Func. Part of acceptance criteria.

game/functions/systems/skills/global/fn_skills_parseTreeCfg.sqf
veteran29: It does not, matter of habit.

game/functions/systems/skills/client/fn_skills_getSkillPoints.sqf
veteran29: Current amount of skill points that player can spend.

game/functions/systems/skills/client/active/fn_skills_active_openSkillWheel.sqf
veteran29: Not a concern for this PR, it's a placeholder UI.

game/functions/systems/skills/client/active/fn_skills_active_preInit.sqf
veteran29: Not a concern for this PR, it's a placeholder UI. Arma modded keybindings can only be used in Mods.

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
veteran29: This is not a concern.

Following applies due to `_known` check:
> Only apply newly learned skills

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
veteran29: The skillTrees hashmap is identical on every client by default and is kinda "locally global", properties that are not coming from config but are used locally are marked with `_`.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
veteran29: `db_get` always returns a HashMap.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
veteran29: If not load it here then where?

The slow path is done only once and it will always happen when a player is joining the server.

The idea behind the cache is that the player can't play on multiple servers at once, so we can re-use the data without polling the DB all the time.

game/functions/systems/skills/server/fn_skills_dbSave.sqf
veteran29: `dbGet` L26

game/functions/systems/skills/server/fn_skills_dbGet.sqf
Spoffy: I'd load it on player join, before anything else happens, so that data is all there and waiting to be used.

In terms of cache - my point was more that I think the behaviour doesn't match the function name. I agree we should be caching it and not re-accessing the DB 🙂

game/functions/systems/skills/server/fn_skills_dbGet.sqf
Spoffy: Cool - you're absolutely right.

game/functions/systems/skills/server/fn_skills_dbSave.sqf
Spoffy: Aha, got it, it's a side-effect of the dbGet, I see that now. Thanks!

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
Spoffy: Ah, got it. I think we need to document that somewhere, maybe in README.md in the skills system directory?

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
Spoffy: Right, I didn't see that check! Thanks :)

game/functions/systems/skills/server/fn_skills_dbGet.sqf
gnalvesteffer: Would recommend naming this `_skillsData`. `_hash` just tells me the datatype (or really it should be `_hashMap` if you want to be accurate). Alternatively `_skillsHashMap`.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
gnalvesteffer: I obviously don't have all the context here, but I do agree with @Spoffy regarding loading all the data upfront from the DB when a player joins, and then caching it, where feasible. Potentially long-running, blocking operations would introduce hitching during play, which is undesirable.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
...
+    Public: No
+
+    Description:
+        Get player skill data from DB or local cache.
+
+    Parameter(s):
+        _player - Player to get the skill data for [OBJECT]
+
+    Returns:
+        Something [BOOL]
```
gnalvesteffer: Is this Return documentation accurate? Looks like it returns the skills data for the player (hashmap).

game/functions/systems/skills/server/fn_skills_dbGet.sqf
gnalvesteffer: Also with that in mind ^, this function could be renamed to `cacheGet` or similar.

veteran29: > I'd load it on player join, before anything else happens, so that data is all there and waiting to be used.

Doesn't change much as first thing player does is call for this data == load it.

> I do agree with @Spoffy regarding loading all the data upfront from the DB when a player joins, and then caching it, where feasible. Potentially long-running, blocking operations would introduce hitching during play, which is undesirable.

It doesn't matter when we load the data if it's synchronous, loading it early during player connection will still freeze the server, unless we want to completely refactor the DB handling and make it async.


Anyways, is this good?: 0f7fa47d821f937268c97f9065482bfe15cefe9a

game/functions/systems/skills/server/network/fn_skills_handle_skillLearnRequest.sqf
veteran29: Not sure if we need it but added.

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
veteran29: Not relevant anymore, the list of already applied skills will be kept in an array.

game/functions/systems/skills/server/fn_skills_dbGet.sqf
gnalvesteffer: It should be async then, but it may or may not be such a big deal, depending on how frequently players join and the max player count. To your point, it wouldn’t matter that much if we load on join or whenever needed, since the server would freeze all the same to the clients. Though loading all data upfront on join would reduce the amount of freezing throughout a playthrough (given we need to pull more data).

game/functions/systems/skills/client/network/fn_skills_receiveSkillsData.sqf
veteran29: Added handling of forgotten skills with skill respec.

game/functions/systems/skills/client/active/fn_skills_active_preInit.sqf
veteran29: Added handling of forgotten skills with skill respec.

Spoffy (approved):

game/functions/systems/skills/server/fn_skills_dbGet.sqf
```diff
@@ -0,0 +1,32 @@
+/*
+    File: fn_skills_dbSave.sqf
```
Spoffy: Header needs an update :)

Spoffy (approved):

game/functions/systems/skills/server/fn_skills_dbGet.sqf
veteran29:
```suggestion
    File: fn_skills_dbGet.sqf
```

PR #17, opened 2023-02-13
feat: Respawn System

gnalvesteffer:
**Description:**
Implements a respawn system according to the [ClickUp ticket](https://app.clickup.com/t/30zwkcc). I assume this is placeholder or a first-pass for a more in-depth system once more of the gamemode is implemented.

Additional efforts outside the ticket include:
- Screen fade transition on death
- Smarter spawn positioning that tries to put the player closer to the majority of their team, ignoring lonewolves and encouraging teamwork.
- Player's direction is oriented towards the group when respawned, giving them a natural guidance on where to go.

**Preview:**
https://medal.tv/games/arma-3/clips/V1GbCnqRlspdd/d1337ZO6XT7c?invite=cr-MSxNOG0sMTYyMzA2MjA1LA

**Acceptance Criteria**
- [x] When a player is killed, they should respawn within 300 to 500m from the team. 
- [x] The direction they spawn in should be random
- [x] They shouldn't spawn in within 100m of an enemy, unless a number of other attempts at spawning them in have been unsuccessful.
- [x] They shouldn't spawn in water.
- [ ] The enemy should no longer be aware of the respawned player **_(I'm not familiar with this, shouldn't the enemy already be unaware of a respawned player?)_**


paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -0,0 +1,55 @@
...
+
+if (isNull _group) exitWith {
+    ["ERROR", format ["Expected _group to be defined. Received _group: %1", _group]] call para_g_fnc_log;
+};
+
+private _unitsInGroup = units _group;
+private _memberCountToMembers = createHashMap;
+{
+    private _unit = _x;
+    private _nearGroupMembers = (_unit nearEntities ["AllVehicles", LONEWOLF_DISTANCE]) select { group _x == _group && _x call _groupMemberPredicate};
```
Spoffy: We can rewrite this in a couple of ways, that'll make it a bit faster, particularly under edge cases. It's not really necessary given how rarely this is called, and the size of the group, but wanted to share in case you'd find it useful!

1) Use "Man" instead of "AllVehicles"
```sqf
_unit nearEntities ["Man", LONEWOLF_DISTANCE]
```
Will save you a little bit of filtering, and dealing with any random vehicles.

2)
```sqf
private _nearGroupMembers = _unitsInGroup inAreaArray  [getPos _unit, LONEWOLF_DISTANCE, LONEWOLF_DISTANCE] select _groupMemberPredicate
```
This has a few advantages: 
* Only loops through group members, so very tightly bounded.
* Avoids `call` inside `select`. The overhead of the extra call on every loop iteration is quite high, whereas using the predicate directly avoids this.

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -0,0 +1,55 @@
...
+    if (count (_nearGroupMembers select { _x != _unit }) > 0) then {
+        _memberCountToMembers set [count _nearGroupMembers, _nearGroupMembers];
+    };
```
Spoffy: Could always just track the highest value in a couple of variables, `private _largestGroup`, `private _largestGroupUnits`, to save yourself the hassle of finding the max later on :)

Do we need to do the `select` here? I think even if the `_unit` is included in the count, all we care about is the largest group, and it means all the counts will just be 1 larger?

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -0,0 +1,55 @@
...
+} forEach _unitsInGroup;
+
+if (count _memberCountToMembers == 0) exitWith {
+    [0, 0, 0];
+};
+
+private _majorityUnits = _memberCountToMembers get (selectMax (keys _memberCountToMembers));
```
Spoffy: Not seen this before, interesting use of the hashmap keys :)

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -0,0 +1,55 @@
...
+private _positionSum = [0, 0, 0];
+private _unitCountSum = 0;
+{
+    private _unit = _x;
+    private _unitPosition = getPos _unit;
+    _positionSum = _positionSum vectorAdd [_unitPosition#0, _unitPosition#1, 0];
```
Spoffy: Could just zero out the 3rd parameter at the end :)

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -0,0 +1,55 @@
...
+    _unitCountSum = _unitCountSum + 1;
```
Spoffy: Shouldn't need to count this, think we can just do `count _majorityUnits` 🙂

game/functions/systems/respawn/server/fn_respawn_onPlayerKilled.sqf
```diff
@@ -0,0 +1,25 @@
...
+        NONE
+
+    Example(s):
+        See `CfgRespawnTemplates/vgm_respawn`.
+*/
+
+params ["_oldUnit", "_killer", "_respawnType", "_respawnDelay"];
+
+sleep 1;
+[0, "WHITE", 3, 1] spawn BIS_fnc_fadeEffect;
```
Spoffy, line 25: I really really love that you've added this.

game/functions/systems/respawn/server/fn_respawn_getInitialSpawnPointMarkerTransform.sqf
```diff
@@ -0,0 +1,28 @@
...
+
+    Example(s):
+        _initialSpawnTransform = call vgm_s_fnc_respawn_getInitialSpawnPointMarkerTransform;
+*/
+
+#define MARKER_NAME "initial_spawn_point"
+
+private _markerPositionASL = AGLToASL (getMarkerPos MARKER_NAME);
+if (_markerPositionASL isEqualTo [0, 0, 0]) exitWith {
+    [getPosASL player, direction player]; // TODO: remove before merge -- this is only for testing!!!
```
Spoffy: Maybe leave this this in as a default behaviour, add an error log message *and* make it do a global `hint`, as this is ultimately a mission configuration error.

Just in case we forget to add or rename this marker and get confused :D

My other suggestion would be to throw an exception or doing something very loud to make the user aware, but Arma doesn't really have anything like that, except for deliberately causing script errors :(

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -0,0 +1,76 @@
...
+private _groupPositionAGL = [group _unit, { alive _this }] call para_g_fnc_get_group_majority_position;
+if (_groupPositionAGL isEqualTo [0, 0, 0]) exitWith {
+    _safeSpawnTransform;
+};
+
+private _enemySides = ([side _unit] call BIS_fnc_enemySides) createHashMapFromArray [];
+_groupPositionASL = AGLToASL _groupPositionAGL;
+
+for "_searchAttempt" from 1 to MAX_SEARCH_ATTEMPTS do {
+    private _safePosition = [_groupPositionASL, _minDistanceFromTeam, _maxDistanceFromTeam, 5, 0, 30, 0, [], [[0, 0], [0, 0]]] call BIS_fnc_findSafePos;
```
Spoffy, line 76: Generally, it's worth avoiding BIS functions because they're sloooooow. But I don't think this is an issue given the specificness of this function, and how rarely this will be called :)

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -0,0 +1,76 @@
...
+};
+
+private _safeSpawnTransform = call vgm_s_fnc_respawn_getInitialSpawnPointMarkerTransform; // fallback spawn position if a safe one can't be found near teammates
+private _groupPositionAGL = [group _unit, { alive _this }] call para_g_fnc_get_group_majority_position;
+if (_groupPositionAGL isEqualTo [0, 0, 0]) exitWith {
+    _safeSpawnTransform;
+};
+
+private _enemySides = ([side _unit] call BIS_fnc_enemySides) createHashMapFromArray [];
+_groupPositionASL = AGLToASL _groupPositionAGL;
```
Spoffy: Think this wants a `private` before it :)

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -0,0 +1,76 @@
...
+
+for "_searchAttempt" from 1 to MAX_SEARCH_ATTEMPTS do {
+    private _safePosition = [_groupPositionASL, _minDistanceFromTeam, _maxDistanceFromTeam, 5, 0, 30, 0, [], [[0, 0], [0, 0]]] call BIS_fnc_findSafePos;
+    if (!(_safePosition isEqualTo [0, 0])) then {
+        _safePosition = AGLToASL [_safePosition#0, _safePosition#1, 0];
+        private _totalNearbyEnemies = { side _x in _enemySides } count (_safePosition nearEntities ["AllVehicles", _enemyAvoidanceDistance]);
```
Spoffy, line 79: `CAManBase` would be better than `AllVehicles` here, as that returns only Units ,:)

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -0,0 +1,76 @@
...
+        private _totalNearbyFriendlies = { !(side _x in _enemySides) } count (_safePosition nearEntities ["AllVehicles", _minDistanceFromTeam]);
```
Spoffy: Same again about `CAManBase`

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -0,0 +1,76 @@
...
+    ["ERROR", format ["Expected _maxDistanceFromTeam to be greater than 0. Received _maxDistanceFromTeam: %1", _maxDistanceFromTeam]] call para_g_fnc_log;
+};
+if (_maxDistanceFromTeam <= _minDistanceFromTeam) exitWith {
+    ["ERROR", format ["Expected _maxDistanceFromTeam to be greater than or equal to _minDistanceFromTeam. Received _minDistanceFromTeam: %1, _maxDistanceFromTeam: %2", _minDistanceFromTeam, _maxDistanceFromTeam]] call para_g_fnc_log;
+};
+if (_enemyAvoidanceDistance < 0) exitWith {
+    ["ERROR", format ["Expected _enemyAvoidanceDistance to be greater than 0. Received _enemyAvoidanceDistance: %1", _enemyAvoidanceDistance]] call para_g_fnc_log;
+};
+
+private _safeSpawnTransform = call vgm_s_fnc_respawn_getInitialSpawnPointMarkerTransform; // fallback spawn position if a safe one can't be found near teammates
```
Spoffy: I've rethought this - let's spawn on top of a random teammate if there's no good spawn position. It disrupts the game flow less. If we spawn a player back at base, they've not got a way back to the mission.

Only just thought of this 😅

game/mission/description.ext
```diff
@@ -52,6 +47,8 @@ enableDebugConsole[] = {
     "76561197974575839",
     //Aceb
     "76561198022706949",
+    //Xorberax
+    "76561198008065973",
```
Spoffy, line 51: 👍

game/mission/respawn.hpp
```diff
@@ -0,0 +1,15 @@
...
+        onPlayerRespawn = "functions\systems\respawn\server\fn_respawn_onPlayerRespawn.sqf";
+    };
+};
+
+respawn = "INSTANT";
+respawnButton = 1;
+respawnDialog = 0;
+respawnDelay = 5;
+respawnOnStart = -1;
+respawnTemplates[] = { "vgm_respawn" };
```
Spoffy, line 15: Overwriting the respawn system entirely like this feels elegant to me 👍

Spoffy (changes requested):
Really cool work - love the attention to detail with the fade effects, and error/edge case handling. 

Left a bunch of comments, but nothing major - mostly just performance or style suggestions. 🙂

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
gnalvesteffer: Oh that's awesome, I didn't know about `inAreaArray`!

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
gnalvesteffer: That sounds right. Also tracking this with just those two variables would be more memory efficient as well.

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
gnalvesteffer: Actually I just realized that the zero is redundant. Will remove.

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
gnalvesteffer: Yup, and it's even simpler now with your other proposed change of tracking the largest group with two variables, one of which is to keep track of the largest group size.

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
gnalvesteffer: Probably just an ARMA-ism, but is there a reason to use `CAManBase` over just `Man`? I've seen and done this in the past a long time ago but forgot why.

gnalvesteffer: Also I'm checking AllVehicles so that the player isn't spawned near an enemy vehicle. `nearEntities` doesn't check for units in vehicles. Let me know if I should keep this as-is or move forward with only checking for CAManBase, or if you have an alternative solution.

image

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
gnalvesteffer: With regards to your other comment about using `CAManBase` for enemies, I think it's fine to replace the friendly unit check however, since it might be kinda silly to try to spawn a player near a friendly vehicle that could be driving off, leaving the player stranded, and also putting a bit of a mental toll on other players with having to (sub)consciously stop their vehicle or search for their teammate when they die. It may be disruptive to the overall team.

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
gnalvesteffer: Agreed.

gnalvesteffer: Also what should we do if the player is the only one left in their group, either because everyone else died and is in the middle of respawning, or because they're the only one on the server? Should they just respawn somewhere around their place of death? Not opposed to that, since it would be respectful of their time to keep them near the action and not have to trek across country from base. At least as a placeholder for now until more of the mission is fleshed out and another approach makes more sense.

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
Spoffy: InareaArray is one of those absolute A-tier commands for anything area based. I think nearly every algorithm I've put in-game for area stuff uses it now 😆

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
Spoffy: Ah, good thinking on the enemy vehicles, that didn't occur to me. I think leave it for now then :)

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
Spoffy: The vehicle edge case is an interesting one. I'm not strongly feeling one way or another on this. I'm not particularly fussed about including vehicles in this check or not - I can thinks of reasons for both behaviours.

Based on that, I wonder if we should just make this into `units group _unitToRespawn inAreaArray []` too, as it'll make the check a little more performance friendly, and include vehicles?

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
Spoffy: Yay, edge cases! 😆. If everyone has died, it's game over, at least under the current designs. But to cover our edge cases... yeah, somewhere around their place of death makes a lot of sense, as it behaves sensibly in all situations, even if that particulary scenario is hit through a bug, or the mission fails to end (for example).

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
gnalvesteffer: Oops, good catch!

game/functions/systems/respawn/server/fn_respawn_findSafeSpawnTransformNearTeam.sqf
gnalvesteffer: Agreed

game/functions/systems/respawn/server/fn_respawn_getInitialSpawnPointMarkerTransform.sqf
gnalvesteffer: Added a hint, error log, and systemChat:
image

game/functions/systems/respawn/server/fn_respawn_findFallbackSpawnTransform.sqf
```diff
@@ -0,0 +1,77 @@
...
+    ["ERROR", format ["Expected _maxDistanceFromSelf to be greater than or equal to _minDistanceFromSelf. Received _minDistanceFromSelf: %1, _maxDistanceFromSelf: %2", _minDistanceFromSelf, _maxDistanceFromSelf]] call para_g_fnc_log;
+};
+
+private _unitGroup = group _unit;
+private _fallbackSpawnUnitPositionASL = getPosASL (selectRandom ((units _unitGroup) select { alive _x && _x != _unit }));
+if (isNil { _fallbackSpawnUnitPositionASL }) then { // no teammates to choose from, use self position
+    _fallbackSpawnUnitPositionASL = getPosASL _unit;
+};
+
+private _fallbackSpawnPositionASL = AGLToASL (([_fallbackSpawnUnitPositionASL, _minDistanceFromTeam, _maxDistanceFromTeam, 5, 0, 30, 0, [], [[0, 0], [0, 0]]] call BIS_fnc_findSafePos) vectorAdd [0, 0, 0]);
```
Spoffy, line 65: What does the vectorAdd at the end of this line do? :)

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
```diff
@@ -8,7 +8,7 @@
 
     Parameter(s):
         _group - The group to get the average position of [GROUP]
-        _groupMemberPredicate - The condition that must be satisfied to count as a nearby group member [CODE]. Optional.
+        _groupMemberPredicate - The condition that must be satisfied to count as a nearby group member. [CODE]. Optional.
```
Spoffy: It'd be more with our usual convention I think to pass the predicate in this form:
```sqf
[
    [_arg1, _arg2],
    {
        // Predicate code
    }
]
```
As that's the format we typically use for callbacks and event handlers, etc.

Spoffy: Usually we then do something like this:
```sqf
if (_predicate isEqualType {}) then {
    _predicate = [[], _predicate]
};
```
To standardise the format, and allow only a code block to be passed :)

Spoffy (commented):
Couple of minor bits, but looking great 👍

game/functions/systems/respawn/server/fn_respawn_findFallbackSpawnTransform.sqf
gnalvesteffer: BIS_fnc_findSafePos returns a 2D AGL position, so my vector add is just a trick to turn it into a 3D position with the altitude being 0, so I can convert it to ASL.

paradigm/global/functions/helpers/fn_get_group_majority_position.sqf
gnalvesteffer: Oh yea that's a lot better since it groups the predicate with its args when calling the function, will change this.

gnalvesteffer: Might be something we can extract to a function later on if we ever get to a point where we're frequently performing this predicate type sanitization.

gnalvesteffer:
@Spoffy Ready for review

Spoffy (approved):

PR #28, opened 2023-05-20
Skilltree UI/backend updates

veteran29:
Updates for adjusted design docs:

- [x] The opposite skill in this level (Tier 1) will lock when the other is chosen as these particular skills are mutually exclusive
- [x] If the player lacks the required skill points to purchase a skill, the plus box will be grayed out. Hovering over will display (“Not enough skill points”
- [x] When a player can invest into a skill, it will show a “+” in the corner. Clicking that plus spends the appropriate number of skill points, adds the skill to the character, and replaces the plus with the check mark.
- [x] When a skill is in a locked tier level, the button on the top right is hidden. The skill can still be clicked on to view its description, though it will clarify that it is locked.
- [x] ~~Certain tiers (rows) will be locked off if a player hasn’t spent enough skill points in this tree. Hovering over the lock will display what is needed to unlock the tier (EX: “2 more skill points needed in the lower tier”)~~ **Edit: Implemented only the "padlock icon" part for now, skill points per tier unlock will need more complex backend changes**
- [x] Show the amount of invested/maximum skill points in the skilltree tree


game/functions/core/client/displays/fn_displaySkills.sqf
```diff
@@ -58,31 +64,98 @@ switch _mode do {
...
+
+        {
+            private _skillTreePath = parseSimpleArray (_ctrlSkills tvData _x);
+            private _skillTree = _skillTreePath call vgm_g_fnc_skills_getByPath;
+
+            private _skillTreePoints = _skillTree call vgm_g_fnc_skills_getTreeSkillPoints;
+            private _label = format ["%1 (%2/%3)", _skillTree get "displayName", _skillTreePoints, _skillTree get "skillPointsMax"];
+
+            _ctrlSkills tvSetText [_x, _label];
+        } forEach (_ctrlSkills getVariable "vgm_skillsListPaths");
```
Spoffy: Is everything in skillsListPaths a skill tree path? If so, it might be worth tweaking the naming to `vgm_skillsListTreePaths` or similar

game/functions/core/client/displays/fn_displaySkills.sqf
```diff
@@ -58,31 +64,98 @@ switch _mode do {
...
+            _ctrlDescription ctrlSetStructuredText parseText (_currentSkill get "description");
+
+            if (_currentSkill call vgm_g_fnc_skills_isKnown) exitWith {
+                _ctrlUnlock ctrlSetText localize "STR_VGM_SKILLS_UI_KNOWN";
+                _ctrlUnlock ctrlEnable false;
+            };
+
+            _ctrlUnlock ctrlSetText format [localize "STR_VGM_SKILLS_UI_UNLOCK", _skill get "cost"];
+            _ctrlUnlock ctrlEnable ([player, _currentSkill] call vgm_g_fnc_skills_canLearn);
+            _ctrlUnlock setVariable ["vgm_params", [_currentSkill]];
```
Spoffy: What's vgm_params? I'm not sure I understand what it is from context alone :)

game/functions/core/client/displays/fn_displaySkills.sqf
```diff
@@ -58,31 +64,98 @@ switch _mode do {
...
+        if (_currentSkillTre isNotEqualTo createHashMap) exitWith {
+
+            _ctrlDescriptionTitle ctrlSetText (format [localize "STR_VGM_SKILLS_UI_SKILL_TREE", _currentSkillTre get "displayName"]);
+            _ctrlDescription ctrlSetStructuredText parseText (_currentSkillTre get "description");
+
+            _ctrlUnlock ctrlShow false;
+            _ctrlUnlock ctrlEnable false;
+            _ctrlUnlock ctrlSetText "";
+        };
+    };
```
Spoffy: Minor typo here in `currentSkillTre`, I think it should be `currentSkillTree`

Spoffy (commented):
Looks awesome - thanks for handling all this UI backend. There's no shortage of code, and it doesn't feel like it's always straightforward code either.

Left a couple of questions. My only overall comment is on whether we need to rename `_skillTree get "skills"` to `_skillTree get "skillTiers"`, since they seem to always be tiers now?

game/functions/core/client/displays/fn_displaySkills.sqf
veteran29: It's a list of tree view item paths, not skill tree paths so it can get confusing a bit here. Renamed to `vgm_skillsListTvPaths` to use same terminology as SQF commands that do stuff with tree view controls.

game/functions/core/client/displays/fn_displaySkills.sqf
veteran29: I've been using it in other UIs to as way to transfer all needed params to event handlers and be able to conveniently do `params` on them, in case of this display it contains only the `_skill` to I've renamed it to `vgm_skill`.

veteran29:
> My only overall comment is on whether we need to rename _skillTree get "skills" to _skillTree get "skillTiers", since they seem to always be tiers now?

We could but I would rather do it in another PR.

Spoffy (approved):

PR #32, opened 2023-06-02
Leveling system, logging tweaks

veteran29:
Simple levelling system, players can gain experience, if the experience is over/equals the current level threshold they "complete" the level and gain the next one. Provides events for hooking by other components.

Additionally, I've done some cleanup of the logging in the `skills` component so it uses our gamemode logging functions.

I've used a similar approach with handling player init as in the `skills` component but I've used events instead of remote executing the functions, I think I like it this way more as it does not end up in so many functions (wanted to do it like this initially but the routing table based event system was not usable for that use case).

### DIFF
--- a/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
+++ b/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
@@ -32,7 +32,11 @@ if (_currentLevel >= vgm_g_leveling_maxLvl) exitWith {
 
 private _currentExperience = _levelingData get "experience";
 
+// award the gained XP
 _currentExperience = _currentExperience + _experience;
+// clamp XP to the amount needed to reach max level
+_currentExperience = _currentExperience min (vgm_g_leveling_levelsHashMap get (vgm_g_leveling_maxLvl-1) get "experience");
+
 _levelingData set ["experience", _currentExperience];
 
 private _currentLevelData = vgm_g_leveling_levelsHashMap get _currentLevel;
--- a/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
+++ b/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
@@ -2,7 +2,7 @@
     File: fn_leveling_addExperience.sqf
     Author: Savage Game Design
     Date: 2023-06-01
-    Last Update: 2023-06-11
+    Last Update: 2023-06-23
     Public: No
 
     Description:
@@ -41,7 +41,7 @@ _levelingData set ["experience", _currentExperience];
 
 private _currentLevelData = vgm_g_leveling_levelsHashMap get _currentLevel;
 
-while {_currentExperience >= (_currentLevelData get "experience")} do {
+while {_currentExperience >= (_currentLevelData get "experienceThreshold")} do {
     // advance to next level
     _currentLevel = _currentLevel + 1;
     (format ["Level %3 gained by %1 (%2)", name _player, getPlayerUID _player, _currentLevel]) call vgm_g_fnc_logInfo;
--- a/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
+++ b/game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
@@ -49,7 +49,7 @@ while {_currentExperience >= (_currentLevelData get "experienceThreshold")} do {
     _currentLevelData = vgm_g_leveling_levelsHashMap get _currentLevel;
     _levelingData set ["level", _currentLevel];
 
-    ["vgm_leveling_levelGained", [_player, _currentLevelData]] call para_g_fnc_event_triggerLocal;
+    ["vgm_leveling_levelGained", [_player, _currentLevelData]] call para_g_fnc_event_triggerGlobal;
 };
 
 [_player, _levelingData] call vgm_s_fnc_leveling_dbSave;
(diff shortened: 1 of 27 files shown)


game/functions/systems/leveling/server/fn_leveling_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+    Returns:
+        Nothing
+ */
+
+if (!isServer) exitWith {};
+
+["vgm_leveling_init", {
+    params ["_player"];
+
+    if (owner _player isNotEqualTo remoteExecutedOwner) exitWith {
```
Spoffy, line 23: (_this # 4) for events is `originMachineId`. You could always use that instead of remoteExecutedOwner. Not sure there's a major benefit in this situation though.

Spoffy: This is the exact signature:
```
    [_this param [DATA_PARAMETER_INDEX], _savedParameters, _eventName, _topic, _originMachineId] call _code;
```

game/functions/systems/leveling/server/fn_leveling_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+        (format ["Leveling data request for %1, owner not matching %2 != %3", name _player, owner _player, remoteExecutedOwner]) call vgm_g_fnc_logError;
+    };
+
+    ["DEBUG", format ["Received player leveling init request %1 (%2)", name _player, getPlayerUID _player]] call vgm_g_fnc_log;
+
+    [_player, 0] call vgm_s_fnc_leveling_addExperience;
```
Spoffy, line 30: What does adding 0 experience do for us here? Can we add the reason for its call here as a comment?

game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
```diff
@@ -0,0 +1,59 @@
...
+while {_currentExperience >= (_currentLevelData get "experience")} do {
+    // advance to next level
+    _currentLevel = _currentLevel + 1;
+    (format ["Level %3 gained by %1 (%2)", name _player, getPlayerUID _player, _currentLevel]) call vgm_g_fnc_logInfo;
+
+    _currentLevelData = vgm_g_leveling_levelsHashMap get _currentLevel;
+    _levelingData set ["level", _currentLevel];
+
+    ["vgm_leveling_levelGained", [_player, _currentLevelData]] call para_g_fnc_event_triggerLocal;
+};
```
Spoffy: I think we also need to set "experience" to 0, so that they start building up experience for the next level again.

As is, if they got the 500 xp to reach level 1, I think they'd still be on 500, and need only another 75xp to advance again?

game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
```diff
@@ -0,0 +1,59 @@
...
+
+while {_currentExperience >= (_currentLevelData get "experience")} do {
+    // advance to next level
+    _currentLevel = _currentLevel + 1;
+    (format ["Level %3 gained by %1 (%2)", name _player, getPlayerUID _player, _currentLevel]) call vgm_g_fnc_logInfo;
+
+    _currentLevelData = vgm_g_leveling_levelsHashMap get _currentLevel;
+    _levelingData set ["level", _currentLevel];
+
+    ["vgm_leveling_levelGained", [_player, _currentLevelData]] call para_g_fnc_event_triggerLocal;
```
Spoffy: We might want to make this a global event, or send it to the levelling up player's client, in the future.

Spoffy (changes requested):

game/functions/systems/leveling/server/fn_leveling_preInit.sqf
veteran29: `0.000185456`ms vs `0.000253346`ms and the intent is more clear with `remoteExecutedOwner` so I'll stick with it.

game/functions/systems/leveling/server/fn_leveling_preInit.sqf
veteran29: It's used to trigger level up from 0 to 1 and broadcast the data to client, added comment.

game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
veteran29: I've added `experienceThreshold` key to the Level HashMaps which holds whole sum of experience needed to complete the level.

game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
veteran29: I want this event to be "sync" with this code, ~~I'll add target event below.~~

Used globalEvent

game/functions/systems/leveling/server/fn_leveling_preInit.sqf
Spoffy: Yup, all good! And headless clients aren't a worry here either :)

game/functions/systems/leveling/server/fn_leveling_addExperience.sqf
Spoffy: This makes me realise why the other change in my PR to the event system is... irritating. 😆

Spoffy: I'll revert the change in mine, so all good here.

Spoffy (approved):

PR #34, opened 2023-07-05
Add status effect API

veteran29:
- Simple API that allows us to toggle status effects like `forceWalk` etc. from multiple components. It will be used by stamina and medical systems.

### DIFF
--- a/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
+++ b/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
@@ -28,6 +28,10 @@ params [
     ["_fnc_onChange", nil, [{}]]
 ];
 
+if (isNil "vgm_c_statusEffect_allEffects") then {
+    vgm_c_statusEffect_allEffects = createHashMap;
+};
+
 vgm_c_statusEffect_allEffects set [_name, _fnc_onChange];
 
 nil
--- a/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
+++ b/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
@@ -1,0 +1,33 @@
+/*
+    File: fnc_statusEffect_create.sqf
+    Author: Savage Game Design
+    Date: 2023-07-02
+    Last Update: 2023-07-05
+    Public: Yes
+
+    Description:
+        Create status effect.
+        The effect callback will be applied once first reason is added or last reason is removed.
+
+    Parameter(s):
+        _name - Name of the status effect [STRING]
+        _fnc_onChange - Function to be called when status effect state changes [CODE]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        ["forceWalk", {
+            params ["_unit", "_inEffect"];
+            _unit forceWalk _inEffect;
+        }] call vgm_c_fnc_statusEffect_create;
+ */
+
+params [
+    ["_name", nil, [""]],
+    ["_fnc_onChange", nil, [{}]]
+];
+
+vgm_c_statusEffect_map set [_name, _fnc_onChange];
+
+nil
--- a/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
+++ b/game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
@@ -2,7 +2,7 @@
     File: fnc_statusEffect_create.sqf
     Author: Savage Game Design
     Date: 2023-07-02
-    Last Update: 2023-07-05
+    Last Update: 2023-07-12
     Public: Yes
 
     Description:
@@ -28,6 +28,6 @@ params [
     ["_fnc_onChange", nil, [{}]]
 ];
 
-vgm_c_statusEffect_map set [_name, _fnc_onChange];
+vgm_c_statusEffect_allEffects set [_name, _fnc_onChange];
 
 nil
--- a/game/functions/core/client/statusEffect/fn_statusEffect_get.sqf
+++ b/game/functions/core/client/statusEffect/fn_statusEffect_get.sqf
@@ -1,0 +1,29 @@
+/*
+    File: fnc_statusEffect_get.sqf
+    Author: Savage Game Design
+    Date: 2023-07-03
+    Last Update: 2023-07-05
+    Public: Yes
+
+    Description:
+        Get status effect state from an unit.
+
+    Parameter(s):
+        _unit - Unit to get the status effect from [OBJECT]
+        _effect - Status effect name [STRING]
+
+    Returns:
+        Status effect state [BOOL]
+
+    Example(s):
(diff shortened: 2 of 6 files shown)


game/functions/core/client/statusEffect/fn_statusEffect_preInit.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+    Description:
+        Status effect client preInit.
+ */
+
+if (!hasInterface) exitWith {};
+
+vgm_c_statusEffect_map = createHashMap;
+
+["forceWalk", {
```
cerebral-314, line 14: Are we initializing/creating all future status effects here?

veteran29: No, only the (common) ones that are related to built-in commands. For example something like "bleeding" related to the medical system should be handled in its own component.

game/functions/core/client/statusEffect/fn_statusEffect_create.sqf
```diff
@@ -0,0 +1,33 @@
...
+            _unit forceWalk _inEffect;
+        }] call vgm_c_fnc_statusEffect_create;
+ */
+
+params [
+    ["_name", nil, [""]],
+    ["_fnc_onChange", nil, [{}]]
+];
+
+vgm_c_statusEffect_map set [_name, _fnc_onChange];
```
Spoffy: Maybe we could change `vgm_c_statusEffect_map` to `vgm_c_statusEffect_allEffects`? 

Makes it a little clearer what it is, but it also helps with it having the same name as the unit variable, which tripped me up when initially reading the code :)

game/functions/core/client/statusEffect/fn_statusEffect_get.sqf
```diff
@@ -0,0 +1,29 @@
...
+    Example(s):
+        [player, "forceWalk"] call vgm_c_fnc_statusEffect_get
+ */
+
+params [
+    ["_unit", objNull, [objNull]],
+    ["_effect", "", [""]]
+];
+
+private _effectMap = _unit getVariable ["vgm_c_statusEffect_map", createHashMap];
```
Spoffy: Similarly, maybe this one could become `vgm_c_statusEffect_currentEffects`? Or `vgm_c_statusEffect_currentEffectsMap`?

game/functions/core/client/statusEffect/fn_statusEffect_preInit.sqf
```diff
@@ -0,0 +1,24 @@
...
+    params ["_unit", "_inEffect"];
+    _unit forceWalk _inEffect;
+}] call vgm_c_fnc_statusEffect_create;
+
+["forceJog", {
+    params ["_unit", "_inEffect"];
+    _unit allowSprint !_inEffect;
+}] call vgm_c_fnc_statusEffect_create;
```
Spoffy: Potential future race condition here if we have other systems register their own effects. If they register in `preInit`, `vgm_c_statusEffect_map` might not exist.

If they register in `postInit`, they might be registered after another system tries to apply them.

For now, maybe we could make `vgm_c_fnc_statusEffect_create` create `vgm_c_statusEffect_map` if it doesn't exist?

game/functions/core/client/statusEffect/fn_statusEffect_remove.sqf
```diff
@@ -0,0 +1,56 @@
...
+
+if (!(_effect in vgm_c_statusEffect_map)) exitWith {
+    format ["Invalid status effect, available values: %1", keys vgm_c_statusEffect_map] call vgm_g_fnc_logError;
+};
+
+private _effectsMap = _unit getVariable "vgm_c_statusEffect_map";
+if (isNil "_effectsMap") then {
+    _effectsMap = createHashMap;
+    _unit setVariable ["vgm_c_statusEffect_map", _effectsMap];
+};
```
Spoffy: What do you think to just returning if the map doesn't exist? As none of the remaining code can trigger anyway (no effect).

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
```diff
@@ -0,0 +1,54 @@
...
+if (isNil "_effectsMap") then {
+    _effectsMap = createHashMap;
+    _unit setVariable ["vgm_c_statusEffect_map", _effectsMap];
+};
+
+format ["Adding status effect reason: %1 | %2", _effect, _reason] call vgm_g_fnc_logInfo;
+
+private _reasonList = _effectsMap getOrDefault [_effect, [], true];
+
+if (_reason in _reasonList) exitWith {};
```
Spoffy, line 54: For each effect, I'd probably have used a hashmap/set of reasons, to prevent duplicates.

I'm not suggesting we change it, I'm just curious what the motivation was behind using an array? 🙂

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
```diff
@@ -0,0 +1,54 @@
...
+    ["_effect", "", [""]],
+    ["_reason", [""]]
+];
+
+
+if (!(_effect in vgm_c_statusEffect_map)) exitWith {
+    format ["Invalid status effect, available values: %1", keys vgm_c_statusEffect_map] call vgm_g_fnc_logError;
+};
+
+private _effectsMap = _unit getVariable "vgm_c_statusEffect_map";
```
Spoffy: Do we need to wipe status effects when a player dies? As I think this would carry over to their new unit?

Spoffy (changes requested):
Few comments, but I like the way this system works :)

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
veteran29: I just needed a simple data structure to hold few strings so just went with an array. Will refactor to HashMap.

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
veteran29: Yeah, object variables persist on respawn.

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
veteran29: Looking at it again, HashMap has no benefits here. I still need to do the `if (_reason in _reasonList) exitWith {};` check, if I won't, the repeated set of same `reason` will trigger `onChange` callback.

Spoffy: Yup, good point! I don't think there's enough reasons then to bother changing it 🙂

Spoffy (approved):

PR #36, opened 2023-07-17
Add common progress bar API

veteran29:
- Progress bar with callbacks
- Similar to CBA API
- Fix execNextFrame not executing if not arguments passed


game/configs/mission/ui/VGM_RscProgressBar.hpp
```diff
@@ -0,0 +1,61 @@
...
+{
+    idd = -1;
+    onLoad = "uiNamespace setVariable ['vgm_rscProgressBar', _this#0]";
+    duration = 1e10;
+    fadeIn = 0;
+    fadeOut = 0;
+
+    class Controls
+    {
+        class Background: RscText
```
7erra, line 16: use VGM_ctrlStatic instead as base class

game/configs/mission/ui/ctrls.hpp
```diff
@@ -13,6 +13,8 @@ import ctrlButtonPicture;
 import ctrlStaticFrame;
 import ctrlStaticPicture;
 import ctrlMap;
+import ctrlMapEmpty;
```
7erra, line 17: Add new class VGM_ctrlMapEmpty instead of using ctrlMapEmpty directly

game/configs/mission/ui/VGM_RscProgressBar.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+            text = "<TEXT>";
+
+            x = VGM_GRID_MIN_X;
+            y = VGM_GRID_MIN_Y;
+            w = VGM_GRID_MAX_W * VGM_GRID_W;
+            h = VGM_GRID_H * 3;
+        };
+
+        class DrawHandler: ctrlMapEmpty
```
7erra: Add and use VGM_ctrlMapEmpty

game/configs/mission/ui/VGM_RscProgressBar.hpp
```diff
@@ -0,0 +1,61 @@
...
+        {
+            idc = VGM_IDC_PROGRESSBAR_PROGRESSBAR;
+
+            x = VGM_GRID_MIN_X + ((VGM_GRID_MAX_W * 0.5) * VGM_GRID_W) * 0.5;
+            y = VGM_GRID_MIN_Y;
+            w = (VGM_GRID_MAX_W * 0.5) * VGM_GRID_W;
+            h = VGM_GRID_H * 3.5;
+        };
+
+        class Title: VGM_ctrlStatic
```
7erra: Use VGM_ctrlTitle as base class instead

game/configs/mission/ui/VGM_RscProgressBar.hpp
```diff
@@ -0,0 +1,61 @@
...
+        {
+            idc = VGM_IDC_PROGRESSBAR_TITLE;
+            style = ST_CENTER;
+            sizeEx = VGM_FONT_S;
```
7erra: Does this have to be this small? VGM_FONT_M should be the default as smaller text sizes might become unreadable

game/configs/mission/ui/VGM_RscProgressBar.hpp
```diff
@@ -0,0 +1,61 @@
...
+            y = "safezoneY";
+            w = "safezoneWAbs";
+            h = "safezoneH";
+        };
+
+        class ProgressBarBackground: VGM_ctrlStatic
+        {
+            colorBackground[] = {0,0,0,0.25};
+
+            x = VGM_GRID_MIN_X + ((VGM_GRID_MAX_W * 0.5) * VGM_GRID_W) * 0.5;
```
7erra: A note about grids: since this is a HUD which exists next to default Arma 3 UI elements it would be wise to use the GUI_GRID instead to make the appearence uniform. Otherwise a user with a small interface size will have an unusually large progress bar. In that case use Rsc... classes from the default game instead and ignore the other base class comments.

game/functions/functions_client.hpp
```diff
@@ -123,6 +123,7 @@ class vgm_c
         VGM_CLIENT_PATH(\core\client\ui);
         class update_loading_screen {};
         class init_loading_text {};
+        class progressBar {};
```
7erra, line 126: Naming convention: progress_bar

game/configs/mission/ui/VGM_RscProgressBar.hpp
veteran29: Font size seemed reasonable in relation to the HUD elements even on a small window with both:

Small:
image

Large:
image

display sizes.

veteran29: Switched to GUI grid and it's using default font size now.

game/configs/mission/ui/VGM_RscProgressBar.hpp
veteran29: Switched to GUI Grid.

game/functions/functions_client.hpp
veteran29: We're using camelCase in the project, old snake case functions are MikeForce leftovers.

7erra:
lgtm

game/functions/core/client/ui/fn_progressBar.sqf
```diff
@@ -0,0 +1,145 @@
...
+            CLOSE_DISPLAY_PROGRESS;
+
+            [_arguments, _startedAt, _duration, "cancelled"] call _fnc_onFailure;
+
+            false // do not block input
+        };
+
+        true // block input
+    }];
+};
```
Spoffy, line 143: This feels like a pretty heavy-handed way of blocking user input. We might want players to still be able to access other menus while the progress bar is happening (for example), or look around. I think it works for now, but are there other options (such as locking them into an animation?) that offer more flexibility?

Spoffy (approved):

PR #37, opened 2023-08-18
Medical - Add feedback effects

veteran29:
- blood overlay during bleeding - https://cdn.discordapp.com/attachments/992380853546127420/1142107771588849684/Replay_2023-08-18_16-44-03_Trim.mp4
- red radial effect when being hit, same as vanilla

### DIFF
--- a/game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
+++ b/game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
@@ -24,6 +24,12 @@
     -1 call vgm_c_fnc_medical_feedbackBleeding;
 }] call para_g_fnc_event_subscribe;
 
+["vgm_medical_woundAdded", {
+    (_this#0) params ["_unit"];
+    if (_unit != player) exitWith {};
+    [] call vgm_c_fnc_medical_feedbackHit;
+}] call para_g_fnc_event_subscribe;
+
 player addEventHandler ["Killed", {
     -1 call vgm_c_fnc_medical_feedbackBleeding;
 }];
--- a/game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
+++ b/game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
@@ -1,0 +1,29 @@
+/*
+    File: fn_medical_feedbackInit.sqf
+    Author: Savage Game Design
+    Date: 2023-07-24
+    Last Update: 2023-08-18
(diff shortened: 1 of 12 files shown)


game/functions/systems/medical/client/fn_medical_postInit.sqf
```diff
@@ -26,47 +26,6 @@ vgm_c_medical_eh = player addEventHandler ["HandleDamage", {call vgm_c_fnc_medic
 } forEach (allPlayers select {!(_x isKindOf "VirtualMan_F")});
 
 // bleeding status effect
-["bleeding", {
-    params ["_unit", "_inEffect"];
+["bleeding", {call vgm_c_fnc_medical_statusEffectBleeding}] call vgm_c_fnc_statusEffect_create;
```
Spoffy, line 29: 👍 - nice refactor

game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
```diff
@@ -0,0 +1,56 @@
...
+
+["vgm_medical_woundAdded", {
+    (_this#0) params ["_unit"];
+    if (_unit != player) exitWith {};
+    [] call vgm_c_fnc_medical_feedbackHit;
+}] call para_g_fnc_event_subscribe;
+
+player addEventHandler ["Killed", {
+    -1 call vgm_c_fnc_medical_feedbackBleeding;
+}];
```
Spoffy: It feels odd to me that we handle 'Turning off' the bleeding feedback here, when it's controlled by the statusEffectBleeding loop. 

Kind of feels to me like the bleeding should be turned off when the status effect gets disabled, rather than the feedback code deciding? 

I think that's because having the logic controlling the overlay existing in two different places makes it harder to understand how the system works, but also is asking for conflicts later on as we change things?

Spoffy: Another option might be to have the feedback overlay be its own status effect, but that seems like it might be over-engineering if we did that right now :)

Spoffy (changes requested):

game/functions/systems/medical/client/feedback/fn_medical_feedback_init.sqf
veteran29: Handled in status effect code. I agree it will make it easier to reason about this feature if it's in a single place.

Spoffy (approved):

PR #40, opened 2023-08-19
Stamina - Add system #VGM-527

veteran29:
- title

Basic stamina system, no coefficients, no UI, TBD in separate PRs.

### DIFF
--- a/game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
+++ b/game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
@@ -7,13 +7,13 @@
     Public: No
 
     Description:
-        No description added yet.
+        Switch local unit to VGM stamina system, disables vanilla stamina.
 
     Parameter(s):
-        N/A
+        _unit - Unit to be switched to VGM stamina system [OBJECT]
 
     Returns:
-        Something [BOOL]
+        Nothing
 
     Example(s):
         [player] call vgm_c_fnc_stamina_unitInit
--- a/game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
+++ b/game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
@@ -1,8 +1,9 @@
+#include "script_component.inc"
 /*
     File: fn_stamina_unitInit.sqf
     Author: Savage Game Design
     Date: 2023-08-18
-    Last Update: 2023-08-18
+    Last Update: 2023-08-19
     Public: No
 
     Description:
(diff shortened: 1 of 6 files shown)


game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
```diff
@@ -0,0 +1,80 @@
...
+        _unit setVariable ["vgm_stamina_exhausted", true];
+        _unit setVariable ["vgm_stamina_exhaustedUntil", time + 10];
+        [_unit, "forceJog", "stamina"] call vgm_c_fnc_statusEffect_set;
+    };
+    if (_exhausted && {time > _unit getVariable "vgm_stamina_exhaustedUntil"}) then {
+        _unit setVariable ["vgm_stamina_exhausted", false];
+        [_unit, "forceJog", "stamina"] call vgm_c_fnc_statusEffect_remove;
+    };
+
+}, [0, _unit]];
```
Spoffy: Will this persist after respawn? Wondering if it might be worth just replacing `unit` with `player`, and not worrying about any complexity adding/removing the handler. That is, if this'll only ever affect the local player?

game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
```diff
@@ -0,0 +1,80 @@
...
+    // crouched, prone and swimming movement are slower so we need to adjust current speed
+    // this will make the stamina costs roughly the same for all stances
+    // this also allows us to balance them easily by changing the "max" speeds of the stances
+    private _coef = animationState _unit call vgm_c_fnc_stamina_getAnimCoef;
+    _speed = (_speed * _coef) min 6;
+
+    // 100 / 4 = 25s of full speed sprint
+    private _drain = linearConversion [0, MAX_SPEED_STD, _speed, -3, 4, true];
+
+    systemChat str ["drain", _drain];
```
Spoffy: Rogue systemChat, flagging it so we can remove it before merging 🙂

Spoffy (approved):
I think we might need an equivalent of `loadCoef`, so we can allow perks to allow players to carry more gear. Any thoughts on how we could implement that cheaply? 

Maybe a coefficient that this system checks?

Looks great otherwise - love how customisable it is going forwards.

game/functions/systems/stamina/client/fn_stamina_unitInit.sqf
veteran29: Forgot to re-apply after respawn. I do not like to hardcode player variable in such features, it's really easy to make it work on any unit and it allows to for example use scripted agents for testing.

veteran29:
The current implementation does not care about amount of the stuff you carry, we would need to discuss how it should affect the stamina, `loadCoef` can still be used to allow carrying more as it's not related directly to stamina system.

Spoffy:
Sounds good, thanks for clarifying :)

PR #51, opened 2023-09-08
Adds AI using the campaign tracker module

Spoffy:
Adds AI to the mission using a hacked-in copy of the tracker module from the campaign.

Intended only as a temporary solution, until we get the Tour AI spawning implemented.

### DIFF
--- a/game/functions/systems/missions/server/internal/fn_missions_despawnMission.sqf
+++ b/game/functions/systems/missions/server/internal/fn_missions_despawnMission.sqf
@@ -1,0 +1,23 @@
+/*
+    File: fn_missions_despawnMission.sqf
+    Author:
+    Date: 2023-09-08
+    Last Update: 2023-09-08
+    Public: No
+
+    Description:
+        Cleans up entities in the mission. Units, locations, scripts, etc.
+
+    Parameter(s):
+        _mission - Mission to cleanup [HASHMAP]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        [_currentMission] call vgm_s_fnc_missions_despawnMission;
+ */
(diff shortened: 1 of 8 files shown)


game/functions/systems/missions/server/internal/fn_missions_despawnMission.sqf
```diff
@@ -0,0 +1,23 @@
...
+    Returns:
+        Nothing
+
+    Example(s):
+        [_currentMission] call vgm_s_fnc_missions_despawnMission;
+ */
+
+params ["_mission"];
+
+terminate (_mission get "overlord_fsm");
```
veteran29: `terminate` does not work on FSMs, FSMs do not have a "handle" like  scripts. It's an ID (number).

Currently overlord does not have a way to stop it but I'll add FSM variable to do that.
```suggestion
(_mission get "overlord_fsm") setFSMVariable ["_forceStop", true];
```

veteran29 (approved):

game/functions/systems/missions/server/internal/fn_missions_despawnMission.sqf
veteran29: Added required feature in `rMASTER16067`

Spoffy: Perfect, I'll merge this for now and then open another PR with the stop feature.

Thanks!

veteran29: You could add it in this PR, support for stopping the FSM will be in internal build tomorrow so...

PR #54, opened 2023-09-09
Loading - Add component

veteran29:
Add new loading system with API that allows other components to register their own handlers which tell us if the component is ready.

Removes bunch of calls to legacy MF stuff.


game/functions/core/client/loading/fn_loading_postInit.sqf
```diff
@@ -0,0 +1,48 @@
...
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        [parameter] call vgm_X_fnc_component_myFunction
+ */
```
Spoffy: Docs 🙂

game/functions/core/client/loading/fn_loading_postInit.sqf
```diff
@@ -0,0 +1,48 @@
...
+            "%1... %2",
+            localize "STR_LOADING",
+            (_handlers apply {_x get "name"}) joinString endl
+        ] call vgm_c_fnc_loading_setText;
+
+        (1 - (count _handlers / _total)) call BIS_fnc_progressLoadingScreen;
+        uiSleep 0.1;
+    };
+
+    "vgm_loading" call BIS_fnc_endLoadingScreen;
```
Spoffy: Having a bit of difficulty understanding the behaviour here. I read this as:
* Call all handlers to start loading, delete if it's done.

* Update text with something like:
 ```
 Loading.... System1
 System2
 System3
 ```
 
 * Update loading bar with a % based on how many handlers haven't yet returned true.
 
 
Does that sound right?

Spoffy: It's worth adding some kind of fallback to this function. If a script errors, we'll get locked in the loading screen and have to restart Arma I believe.

Spoffy: Can you shoot while on the loading screen? I vaguely recall that being the reason we originally disabled player simulation while loading.

Spoffy (changes requested):

game/functions/core/client/loading/fn_loading_postInit.sqf
veteran29: > Does that sound right?

yes

> Can you shoot while on the loading screen? I vaguely recall that being the reason we originally disabled player simulation while loading.

No, if you `BIS_fnc_startLoadingScreen` during `initFunctions` `postInit` it will be added to the "stack" of loading screens. Control over player avatar is given after all of them are done. I think you needed that as you did not use BI func and fired the loading screen at some later point of execution.

veteran29: > It's worth adding some kind of fallback to this function.

I think we need a timeout, but what should it do? In "production" it should probably kick you from the game telling you that something went wrong. Not sure about dev, maybe kick if no debug console and print an error and close loading screen if player has access to it?

veteran29: Added kick for standard user and error message for admins/debug console users after a 30s timeout.

Spoffy (approved):

PR #55, opened 2023-09-14
Paradigm - Makes netmaps easier / more reliable to clean up

Spoffy:


paradigm/server/functions/netmap/fn_netmap_terminate.sqf
```diff
@@ -32,10 +32,26 @@ if (isNil "_netmapDetails") exitWith {
...
+
+{
+    {
+        private _ownedNetmap = _netmaps get _x;
+        _netmapsToTerminate pushBack _ownedNetmap;
+
+        private _newOwnedNetmapIds = _ownedNetmap get "_netmap" get "ownedNetmaps";
+        _ownedNetmapIds pushBack _newOwnedNetmapIds;
+    } forEach _x
+} forEach _ownedNetmapIds;
```
veteran29: This part feels hard to grasp, appending to the array we're iterating over is quite dirty.

I think function recursion would be preferred for readability. Something in line of (I might've missed something and below might not work):
```sqf
{
    private _ownedNetmap = _netmaps get _x;
    [_ownedNetmap] call para_s_fnc_netmap_terminate;
} forEach (_netmapDetails get "ownedNetmaps");

private _id = _netmapDetails get "id";
_netmaps deleteAt _id;
[_id] remoteExecCall ["para_c_fnc_netmap_terminate", -clientOwner];
```

veteran29 (changes requested):

paradigm/server/functions/netmap/fn_netmap_terminate.sqf
Spoffy: I'm pretty sure this would just work if we went recursive with it.

Only downside is it could be a performance hit if it's dealing with a lot of netmaps terminating at once, and closes the door on bulk-terminating on the client in the future (to reduce number of remoteExec calls)

I don't particularly mind - in this case I chose performance, but I can switch to the recursive solution if you prefer readability. Can use the git history to keep the performance version too for later if needed?

veteran29: If we were creating and deleting netmaps that often this would be an issue I think we would have a design problem. In this case, I prefer readability over the performance approach.

veteran29 (approved):
LGTM

PR #58, opened 2023-09-20
Missions - Add group, change event names

veteran29:
- title


game/functions/systems/missions/client/internal/fn_missions_finishDeploy.sqf
```diff
@@ -29,6 +29,7 @@ private _safeStartPosASL = _defaultStartPosASL findEmptyPosition [1, 20, "CAManB
 private _startPosASL = [AGLtoASL _safeStartPosASL, _defaultStartPosASL] select (_safeStartPosASL isEqualTo []);
 
 player setPosASL _startPosASL;
+[player] joinSilent (_currentMission get "group");
```
Spoffy: What do you think to joining players to the group when they join the mission (attachPlayerToMission function), rather than when deployment finishes on the client?

game/functions/systems/missions/server/fn_missions_endMission.sqf
```diff
@@ -32,9 +32,10 @@ private _missionMemberPlayerIds = (_missionPublic get "players") call para_g_fnc
 private _missionMemberMachineIds = values (_mission get "machineIds");
 
 [] remoteExecCall ["vgm_c_fnc_missions_endMission", _missionMemberMachineIds];
+units (_missionPublic get "group") joinSilent vgm_core_lobbyGroup;
```
Spoffy: Similar to the other comment, I'm wondering if this should belong in `removePlayerFromMission`?

Spoffy (changes requested):

game/functions/systems/missions/server/fn_missions_endMission.sqf
veteran29: Not sure where it should belong as there's a lot of similar functions ;D moved group handling to attach/remove functions.

Spoffy (approved):

PR #83, opened 2023-11-19
Equipment - Filter loadouts in arsenal

veteran29:
- loadout filtering on arsenal close and template loadout loading
- dependent on gear available to player (`vgm_equipment`)

### DIFF
--- a/game/functions/systems/equipment/client/fn_equipment_arsenalInit.sqf
+++ b/game/functions/systems/equipment/client/fn_equipment_arsenalInit.sqf
@@ -3,7 +3,7 @@
     File: fn_equipment_arsenalInit.sqf
     Author: Savage Game Design
     Date: 2023-11-19
-    Last Update: 2023-11-19
+    Last Update: 2023-11-20
     Public: No
 
     Description:
@@ -39,7 +39,7 @@
 
         _display spawn {
             private _loadout = getUnitLoadout player;
-            private _filteredLoadout = [+_loadout] call vgm_c_fnc_equipment_filterLoadout;
+            private _filteredLoadout = [_loadout] call vgm_c_fnc_equipment_filterLoadout;
             if (_filteredLoadout isNotEqualTo _loadout) then {
                 ["showMessage", [_this, localize "STR_A3_RscDisplayArsenal_message_unavailable"]] call BIS_fnc_arsenal;
                 player setUnitLoadout _filteredLoadout;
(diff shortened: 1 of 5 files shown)


game/functions/systems/equipment/client/fn_equipment_arsenalInit.sqf
```diff
@@ -0,0 +1,66 @@
...
+        Nothing
+
+    Example(s):
+        [parameter] call vgm_c_fnc_equipment_arsenalInit
+ */
+
+[true, "arsenalPreOpen", {
+    // Add our special Medical items to "Misc" tab
+    {BIS_fnc_arsenal_data select 24 pushBackUnique _x} forEach ["vn_helper_item_firstaidkit", "vn_helper_item_medikit"];
+}] call BIS_fnc_addScriptedEventHandler;
```
Spoffy, line 25: ❤️

game/functions/systems/equipment/client/fn_equipment_filterLoadout.sqf
```diff
@@ -0,0 +1,103 @@
...
+    {
+        if (count _x == 7) then {
+            _x call _fnc_checkWeapon;
+            continue;
+        };
+        private _itemClass = _x select 0;
+        if (!(_itemClass in _allowedItems)) then {
+            _items set [_forEachIndex, []];
+        };
+        // TODO is it possible to have container in container?
```
Spoffy: Yup, turns out you can have vests and uniforms inside backpacks. Here's what it looks like:
```
[[],[],[],[],[],["vn_c_pack_02",[["vn_o_vest_04",false],["vn_b_uniform_macv_03_17",false]]],"","",[],["vn_b_item_map","","vn_b_item_radio_urc10","vn_b_item_compass_sog","vn_b_item_watch",""]]
```
probably worth addressing now, as players will definitely find this loophole :D

game/functions/systems/equipment/client/fn_equipment_filterLoadout.sqf
Spoffy, line 1: To me, `fn_equipment_filterLoadout` to be a little shennanigany, modifying its input array when the functions are called 'check'.

The backpack one where it literally resizes _this threw me off for a minute, until I realised _this was a reference to the backpack entry in the loadout array :D

As a cheap change, can I propose we rename `fn_checkX` functions to `fn_filterX` or `removeForbiddenFromX`? I think it more strongly implies it's doing some work on the input arrays.

I'd probably prefer something like `_loadout set [_x, _newEntry]` in the filter loops, but I don't think it's worth the time spent making that change.

Spoffy (changes requested):
Overall looks good! 

Couple of proposed tweaks :)

game/functions/systems/equipment/client/fn_equipment_filterLoadout.sqf
veteran29: > I'd probably prefer something like _loadout set [_x, _newEntry] in the filter loops, but I don't think it's worth the time spent making that change.

It's bit hard to follow but it's a lot easier to just modify via ref, that's why I went with it.

I agree on the naming change, the function originally was called `checkGear` and I forgot to change the internal functions.

Spoffy (approved):
Awesome, thanks for making those changes, it definitely felt easier to understand! :)

PR #84, opened 2023-11-19
Skills - Disable actives keybind in Arsenal/Dialogs

veteran29:
- Prevents player from getting stuck in arsenal camera if he presses the keybind in the arsenal


Spoffy (changes requested):
Can we put a comment in why this is needed? :)

veteran29:
> Can we put a comment in why this is needed? :)

Added

Spoffy (approved):

PR #100, opened 2023-11-27
Skills - Add Radio Support config, enable RTO trait

veteran29:
- title
- Adds radio support config, CAS is unavailable until "Heavy Support" skill is taken
- Only Illumination available by default


Spoffy (approved):

Spoffy:
I'd like more than illumination by default with the RTO loadout, but I'll sort that, this is fantastic :)

PR #101, opened 2023-12-01
Carrying - Add system

veteran29:
- title

### DIFF
--- a/game/functions/systems/carry/client/fn_carry_doCarry.sqf
+++ b/game/functions/systems/carry/client/fn_carry_doCarry.sqf
@@ -6,7 +6,7 @@
     Public: No
 
     Description:
-        Make unit carry the target with pickup animations.
+        Make unit carry the target.
 
     Parameter(s):
         _unit   - Unit trying to carry [OBJECT]
--- a/game/functions/systems/carry/client/fn_carry_doCarry.sqf
+++ b/game/functions/systems/carry/client/fn_carry_doCarry.sqf
@@ -2,7 +2,7 @@
     File: fn_carry_canCarry.sqf
     Author: Savage Game Design
     Date: 2023-11-03
-    Last Update: 2023-12-01
+    Last Update: 2023-12-06
     Public: No
 
     Description:
@@ -26,15 +26,6 @@ if (!isNull (_unit getVariable ["vgm_carry_carriedObject", objNull])) exitWith {
 };
 
 format ["Do carry: %1 | %2", _unit, _target] call vgm_g_fnc_logInfo;
-
-_unit setVariable ["vgm_carry_carriedObject", _target, true];
-
-// position the target so anims are somewhat synchronized
-_target attachTo [_unit, [0.22,0.7,0]];
-_target setDir 180;
-
-[_unit, "AcinPknlMstpSnonWnonDnon_AcinPercMrunSnonWnonDnon"] remoteExec ["switchMove", 0];
-[_target, "AinjPfalMstpSnonWrflDnon_carried_Up"] remoteExec ["switchMove", 0];
 
 // request attach from server to prevent race conditions
 [_unit, _target] remoteExec ["vgm_s_fnc_carry_attachRequest", 2];
--- a/game/functions/systems/carry/client/fn_carry_doCarry.sqf
+++ b/game/functions/systems/carry/client/fn_carry_doCarry.sqf
@@ -2,11 +2,11 @@
     File: fn_carry_canCarry.sqf
     Author: Savage Game Design
     Date: 2023-11-03
-    Last Update: 2023-11-03
+    Last Update: 2023-12-01
     Public: No
 
     Description:
-        Check if unit can carry the target.
+        Make unit carry the target with pickup animations.
 
     Parameter(s):
         _unit   - Unit trying to carry [OBJECT]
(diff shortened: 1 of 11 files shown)


game/functions/systems/carry/client/fn_carry_doCarry.sqf
```diff
@@ -0,0 +1,40 @@
...
+format ["Do carry: %1 | %2", _unit, _target] call vgm_g_fnc_logInfo;
+
+_unit setVariable ["vgm_carry_carriedObject", _target, true];
+
+// position the target so anims are somewhat synchronized
+_target attachTo [_unit, [0.22,0.7,0]];
+_target setDir 180;
+
+[_unit, "AcinPknlMstpSnonWnonDnon_AcinPercMrunSnonWnonDnon"] remoteExec ["switchMove", 0];
+[_target, "AinjPfalMstpSnonWrflDnon_carried_Up"] remoteExec ["switchMove", 0];
```
Spoffy: It feels weird to me that we trigger this clientside, before confirming with the server that we're able to do the attach. 

Would it reduce the risk of weird situations cropping up, if we triggered all this on the server, only after the server's confirmed we can attach?

game/functions/systems/carry/server/remoteExec/fn_carry_attachRequest.sqf
```diff
@@ -0,0 +1,52 @@
...
+[_id, {
+    // exploit implicitly passed handler ID to get the unit via netId
+    private _unit = (_id splitString "$" param [1, "-1"]) call BIS_fnc_objectFromNetId;
+    _target = _unit getVariable ["vgm_carry_carriedObject", objNull];
+
+    format ["Delayed attach: %1 | %2", _unit, _target] call vgm_g_fnc_logDebug;
+
+    if (!alive _unit || (lifeState _unit == "INCAPACITATED")) exitWith {
+        format ["Carrier is downed, aborting: %1 | %2", _unit, _target] call vgm_g_fnc_logWarning;
+    };
```
Spoffy: If we abort here, we're already attached because of doCarry, and have variables set. I think we need to roll a bunch of things back?

game/functions/systems/carry/client/fn_carry_doCarry.sqf
Spoffy: (Same for line 30 - where we set the variable)

game/functions/systems/carry/server/remoteExec/fn_carry_attachRequest.sqf
```diff
@@ -0,0 +1,52 @@
...
+
+    _target attachTo [_unit, [0.2, -0.07, -1.1], "spine3"];
+    _target setDir 0;
+
+    _target switchMove "vn_carried_still";
```
Spoffy: Does switchMove need remoteExec'ing here?

game/functions/systems/carry/server/remoteExec/fn_carry_detachRequest.sqf
```diff
@@ -0,0 +1,53 @@
...
+    detach _target;
+
+    if (lifeState _unit != "INCAPACITATED") then {
+        [_unit, ""] remoteExec ["switchMove"];
+    };
+
+    [_target, ([
+        "AmovPpneMstpSnonWnonDnon",
+        "UnconsciousReviveDefault"
+    ] select (lifeState _target == "INCAPACITATED"))] remoteExec ["switchMove"];
```
Spoffy: Do need an `alive` check here too, in case the player being dropped is dead?

Spoffy (changes requested):

game/functions/systems/carry/client/fn_carry_doCarry.sqf
veteran29: According to Robs feedback I've removed the animations and will use short hold actions instead. Everything animation related is now done on the server.

game/functions/systems/carry/server/remoteExec/fn_carry_attachRequest.sqf
veteran29: Dismissing as it's not relevant after removal of animations.

game/functions/systems/carry/server/remoteExec/fn_carry_attachRequest.sqf
veteran29: It did, switchMove needs local target

game/functions/systems/carry/server/remoteExec/fn_carry_detachRequest.sqf
veteran29: It shouldn't be needed, dying unit will use "face down" anim and then ragdoll.

game/functions/systems/carry/client/remoteExec/fn_carry_attachResponse.sqf
```diff
@@ -0,0 +1,58 @@
...
+    Returns:
+        Nothing
+
+    Example(s):
+        [_unit, _target] remoteExec ["vgm_c_fnc_carry_attachResponse", _unit];
+ */
+
+params ["_unit", "_target"];
+
+_unit setVariable ["vgm_carry_carriedObject", _target];
```
Spoffy: Feels like the server should control this, or this should be broadcast over the network. 

Setting it globally on the server in some situations, and locally on just the carrying client in others, feels like it could lead to edge cases where there's a mismatch between clients.

Spoffy (changes requested):

game/functions/systems/carry/client/remoteExec/fn_carry_attachResponse.sqf
veteran29: I agree, this is leftover of the old implementation with animations and is already broadcasted from server so doing it again here is redundant.

Spoffy (approved):

PR #110, opened 2023-12-18
Behaviour Tree - Add component

Spoffy:
Adds an initial version of the behaviour tree framework.

The logic could use a little bit of tidying and simplification but it fundamentally works.


game/functions/systems/behaviour_trees/global/fn_btree_childFinished_action.sqf
```diff
@@ -0,0 +1,27 @@
...
+
+    Returns:
+        The next action to perform:
+            [_nextActionParams, _nextAction] [ARRAY, CODE]
+
+    Example(s):
+        N/A
+ */
+
+#include "..\behaviour_trees.inc"
```
veteran29: Move includes above block comment/header to prevent issues with line count in script errors.

game/functions/systems/behaviour_trees/global/fn_btree_tickGroup.sqf
```diff
@@ -0,0 +1,84 @@
...
+
+if (isNil "_tree") exitWith {};
+
+// These variables are available for ALL CALLED FUNCTIONS.
+// I.e - It should be safe to use them in any non-setup btree functions (actions, etc)
+// However - be careful, as changing them could break execution.
+private _group = _group;
+private _btreeState = _group getVariable "vgm_l_btree_state";
+private _stack = _btreeState get "stack";
+private _blackboard = _btreeState get "blackboard";
```
veteran29: I think we could consider some naming convention for variables like this:
`_g_stack` | `_extern_stack` etc.

would make them easier to spot in the functions that do use them.

veteran29 (approved):
I do not have experience of working with behaviour trees but from quick documentation reads about how they're supposed to work it looks solid, at least basing it on my currently limited knowledge.

game/functions/systems/behaviour_trees/global/fn_btree_childFinished_action.sqf
Spoffy: Thanks for the reminder, I'll sort it :D

game/functions/systems/behaviour_trees/global/fn_btree_tickGroup.sqf
Spoffy: Yeah, I agree. `_extern_` is a pretty good idea honestly, since it plays on the existing semantics from other languages.

`_inherit_`, `_env_` or `_public_`, `_scope_`  might ones I'd suggest too. 

Of all the ones suggested (by both you and me), what do you feel communicates the idea the clearest?

veteran29: `_inherit` or `_extern` I feel like these fit SQF the most.

Spoffy: Of those two, I think I prefer `_extern`, as it more clearly communicates "Not found here", whereas `_inherit` maybe isn't clear unless you've seen it before, so let's go with that 🙂

Spoffy: Pushed changes, to make all those inherited environment variables _extern.

I've also done a slight reorganisation, putting the files in a nested folder.

game/functions/systems/behaviour_trees/runner/global/fn_btree_tickGroup.sqf
```diff
@@ -0,0 +1,84 @@
...
+private _nextAction = ACTION_RUN_CURRENT_NODE;
+private _nextActionParams = [];
+
+// If no current node, start tree execution from the top.
+if (count _extern_stack == 0) then {
+    _nextAction = ACTION_ENTER_NODE;
+    _nextActionParams = [_tree];
+};
+
+[format ["Ticking behaviour tree for group %1 at %2", str _group, serverTime]] call vgm_g_fnc_btree_log;
```
veteran29, line 47: This might spam RPT quite a bit.

veteran29 (approved):
Code LGTM.

There's no code that runs the "tickGroup", will this be added with AI behaviour?

game/functions/systems/behaviour_trees/runner/global/fn_btree_childFinished_decorator.sqf
```diff
@@ -0,0 +1,39 @@
...
+
+    Returns:
+        The next action to perform:
+            [_nextActionParams, _nextAction] [ARRAY, CODE]
+
+    Example(s):
+        N/A
+ */
+
+#include "..\behaviour_trees.inc"
```
veteran29: These still need sorting out.

Spoffy: Thanks for the reminder! I'll push a fix for these shortly and then merge.

Spoffy:
> Code LGTM.
> 
> There's no code that runs the "tickGroup", will this be added with AI behaviour?

Yeah, I'm just getting the functionality reliable for now! We'll likely a slightly tweaked version of the same loop the current behaviour system uses, as that's proven to work fine as AI move across clients. 🙂

game/functions/systems/behaviour_trees/runner/global/fn_btree_tickGroup.sqf
Spoffy: btree_log is special. It saves it in a local variable on the group, rather than going straight to the RPT (especially since diag_log calls are surprisingly expensive!).

The idea is to only access them if you need to debug the AI. I probably should add some mechanism to trim the history on these though to prevent them getting too big and out of hand, and disable them by default.

PR #111, opened 2023-12-20
Mission Objects - Add system #VGM-223

veteran29:
Adds basic framework that allows creation of "virtual" items which are spawned as local objects on the clients participating in the mission. Objects do not exist on the server or clients not in the mission which improves performance.

Provides:
- create function which "registers" the object in the mission
- delete function which removes the virtual item on the server and object instance on the clients
- virtual objects created before mission start will be automatically instanced by the clients
- virtual objects created during the mission need to be sent to the clients, easy API function provided

### DIFF
--- a/game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
+++ b/game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
@@ -6,19 +6,23 @@
     Public: No
 
     Description:
-        No description added yet.
+        Create "real" instances of virtual mission objects.
 
     Parameter(s):
-        N/A
+        _missionObjects - Mission objects [HASHMAP]
 
     Returns:
-        Something [BOOL]
+        Nothing
 
     Example(s):
-        [parameter] call vgm_c_fnc_mission_objects_spawnObjects
+        [_missionObjects] call vgm_c_fnc_mission_objects_spawnObjects
  */
 
 params ["_missionObjects"];
+
+if (remoteExecutedOwner != 2) then {
+    format ["Mission object creation should be called from server to prevent object desync, expected 2 got %1", remoteExecutedOwner] call vgm_g_fnc_logWarning;
+};
 
 format ["Spawning %1 local mission objects", count _missionObjects] call vgm_g_fnc_logInfo;
 
--- a/game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
+++ b/game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
@@ -1,0 +1,37 @@
+/*
+    File: fn_spawnObjects.sqf
+    Author: Savage Game Design
+    Date: 2023-12-20
+    Last Update: 2023-12-20
+    Public: No
+
+    Description:
+        No description added yet.
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        [parameter] call vgm_c_fnc_mission_objects_spawnObjects
+ */
+
+params ["_missionObjects"];
+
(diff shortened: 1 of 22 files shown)


game/functions/systems/missions/server/fn_missions_startMission.sqf
```diff
@@ -39,6 +39,42 @@ if (_missionPublic get "status" isNotEqualTo "CREATED") exitWith {
 
 [_mission] call vgm_s_fnc_director_startMission;
 [_missionPublic get "startPosASL"] call vgm_s_fnc_missions_gameplay_ambient_departHelicopter; // TODO by what and where should this be fired?
+// TODO what should setup objects in the mission?
```
Spoffy, line 42: Probably the director, but not 100% sure right now. No change needed here, just commenting :)

game/functions/systems/mission_objects/server/fn_mission_objects_spawnObjects.sqf
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+params ["_mission", "_missionObjects"];
+private _missionId = _mission get "public" get "id";
+
+// turn object ids into a spawn "chunk"
+if (_missionObjects isEqualType []) then {
+    private _allMissionObjects = vgm_s_mission_objects_data getOrDefault [_missionId, createHashMap];
+    _missionObjects = createHashMapFromArray (_missionObjects select {_x in _allMissionObjects} apply {[_x, _allMissionObjects get _x]});
+};
```
Spoffy, line 29: This looks to generate a hashmap, but clientside spawnObjects looks to expect an array. Do we need `_allMissionsObjects get _x`? Or will an array of IDs suffice?

game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+format ["Spawning %1 local mission objects", count _missionObjects] call vgm_g_fnc_logInfo;
+
+{
+    private _id  = _x;
+    if (_id in vgm_c_mission_objects) then {continue};
+    _y params ["_class", "_position", "_dir", "_fnc_init", "_initParams"];
+
+    private _object = createVehicleLocal [_class, _position, [], 0, "CAN_COLLIDE"];
+    _object setDir _dir;
```
Spoffy: Should we disable simulation or maybe use a simple object? Wondering for desync reasons due to gravity, being knocked, etc

Spoffy (changes requested):
Looks great! Couple of comments, but this looks perfect for both the scouting/hint objects and oddities! 

For the latter, I wonder if it's worth spawning a copy on the server to make sure AI pathfinding doesn't get messed with?

game/functions/systems/mission_objects/server/fn_mission_objects_spawnObjects.sqf
veteran29: Client side expects hashmap in the format this `if` generates. This converts array of IDs into a "chunk" which allows us to send only the objects we want to less network load.

game/functions/systems/mission_objects/client/fn_mission_objects_spawnObjects.sqf
veteran29: Made it so it uses simple objects by default, and if there's an init code it will use object with simulation disabled (simple objects do not support actions)

veteran29:
Objects are spawned also on the server now.

game/functions/systems/mission_objects/server/fn_mission_objects_spawnObjects.sqf
Spoffy: Ah, got it. And yup, I see the line I didm't see before in the clientside script now :D

Spoffy (approved):

PR #112, opened 2024-01-22
Skill Investigate - Add component

veteran29:
- title

Thanks to HorribleGoat for noise texture.


game/functions/systems/skill_investigate/client/script_component.inc
```diff
@@ -0,0 +1,12 @@
...
+
+#define WAVE_SPEED 45
+#define ICON_BASE_SIZE 0.1
+#define ICON_BASE_DIST 10
+#define ICON_ALPHA 0.4
+#define FADE_TIME_COEF 5
+
+#define WAVE1_TIME 0
+#define WAVE2_TIME 0.8
+#define WAVE3_TIME 1.6
```
Spoffy: Are these are the time thresholds, in seconds, when that wave starts showing?

If so, maybe we could use `WAVE1_START_TIME`, as it would make the logic a little easier to understand :)

game/functions/systems/skill_investigate/client/fn_skill_investigate_setListenMode.sqf
```diff
@@ -0,0 +1,58 @@
...
+        } forEach (_noiseSources select {group _x != _ignoredGroup});
+
+        _thisArgs set [0, time + PING_TICK];
+    };
+
+    //----- draw sound sources
+    {
+        private _completed = _x call vgm_c_fnc_skill_investigate_drawSoundWaves;
+        if (_completed) then {vgm_c_skill_investigate_noises deleteAt _forEachIndex}
+    } forEachReversed vgm_c_skill_investigate_noises;
```
Spoffy: Would this simplify to:
```suggestion
    vgm_c_skill_investigate_noises select { !( _x call vgm_c_fnc_skill_investigate_drawSoundWaves) }
```
?

Spoffy (changes requested):

game/functions/systems/skill_investigate/client/fn_skill_investigate_setListenMode.sqf
veteran29: Good catch, it will be a lot faster too. Just needs an assignment as `select` does not modify input array.
```sqf
// Execution Time: 0.0050 ms  |  Cycles: 10000/10000  |  Total Time: 50 ms
test_array = [false, true, true, false, false, true];
{
	private _completed = _x;
	if (_completed) then {test_array deleteAt _forEachIndex};
} forEachReversed test_array;
```
```sqf
// Execution Time: 0.0017 ms  |  Cycles: 10000/10000  |  Total Time: 17 ms
test_array = [false, true, true, false, false, true];

test_array = test_array select {!_x};
```

Spoffy (approved):

PR #113, opened 2024-02-02
Adds compiler, basic nodes and basic test to behaviour trees

Spoffy:
This PR includes:
- Compiler for the behaviour trees
- A (manually run) test case, with an example tree.
- Some fixes for decorators
- Support for services (i.e Nodes that keep running while they're in the stack)
- onTreeAssigned and onTreeUnassigned, to allow nodes to register event handlers.

### DIFF
--- a/game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
+++ b/game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
@@ -2,7 +2,7 @@
     File: fn_btree_compileTree.sqf
     Author: Savage Game Design
     Date: 2024-01-26
-    Last Update: 2024-02-02
+    Last Update: 2024-02-03
     Public: No
 
     Description:
@@ -19,6 +19,10 @@
 
 params ["_tree"];
 
+// Hashmaps prevent identical code running multiple times.
+private _extern_onTreeAssignedCallbacks = createHashMap;
+private _extern_onTreeUnassignedCallbacks = createHashMap;
+
 // Recursive algorithm should be fine, as our behaviour trees are unlikely to ever be more than 10 nodes deep.
 private _fnc_compileNode = {
     params ["_constructor", "_params", "_children"];
@@ -26,7 +30,22 @@ private _fnc_compileNode = {
     private _childNodes = _children apply {_x call _fnc_compileNode};
 
     // By passing built children to the constructor, we enable the constructor to modify them as needed.
-    [createHashMapFromArray _params, _childNodes] call _constructor
+    private _node = [createHashMapFromArray _params, _childNodes] call _constructor;
+
+    if ("onTreeAssigned" in _node) then {
+        _extern_onTreeAssignedCallbacks set [_node get "onTreeAssigned", true];
+    };
+
+    if ("onTreeUnassigned" in _node) then {
+        _extern_onTreeUnassignedCallbacks set [_node get "onTreeUnassigned", true];
+    };
+
+    _node
 };
 
-_tree call _fnc_compileNode
+createHashMapFromArray [
+    ["rootNode", _tree call _fnc_compileNode],
+    ["onTreeAssignedCallbacks", keys _extern_onTreeAssignedCallbacks],
+    ["onTreeUnassignedCallbacks", keys _extern_onTreeUnassignedCallbacks]
+]
+
--- a/game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
+++ b/game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
@@ -1,0 +1,32 @@
+/*
+    File: fn_btree_compileTree.sqf
+    Author: Savage Game Design
+    Date: 2024-01-26
+    Last Update: 2024-02-02
+    Public: No
+
+    Description:
+        Compiles a runnable behaviour tree, from a behaviour tree in array format.
+
+    Parameter(s):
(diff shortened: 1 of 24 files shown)


game/functions/systems/behaviour_trees/nodes/global/fn_btree_decorator_basic.sqf
```diff
@@ -0,0 +1,89 @@
...
+            - If true, run the "onTick" code every tick, while this decorator is running
+              (i.e, any children are running)
+
+        "onTreeAssigned"
+            - Called when the tree is assigned to a group
+            - Useful for setting up event handlers.
+
+        "onTreeUnassigned"
+            - Called when the tree is unassigned from the group or the group is deleted
+            - Useful for cleaning up event handlers.
```
veteran29, line 46: Is it me or are these not implemented?

game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
```diff
@@ -0,0 +1,51 @@
...
+        A compiled behaviour tree, ready for assigning to a group and running [HASHMAP]
+
+    Example(s):
+ */
+
+params ["_tree"];
+
+// Hashmaps prevent identical code running multiple times.
+private _extern_onTreeAssignedCallbacks = createHashMap;
+private _extern_onTreeUnassignedCallbacks = createHashMap;
```
veteran29, line 24: If the callbacks are on the node level why we do not allow duplicate code?

game/functions/systems/behaviour_trees/compiler/global/fn_btree_runBasicTest.sqf
```diff
@@ -0,0 +1,89 @@
...
+
+private _testGroup = createGroup civilian;
+
+[_testGroup, _compiledTree] call vgm_g_fnc_btree_setTree;
+
+[_testGroup] call vgm_g_fnc_btree_tickGroup;
+[_testGroup] call vgm_g_fnc_btree_tickGroup;
+
+// Change the tree so the UnassignTree callbacks get called.
+private _alternateTree = [ACTION(basic), []] call vgm_g_fnc_btree_compileTree;
```
veteran29: I think this should be:
```suggestion
private _alternateTree = [[ACTION(basic), []]] call vgm_g_fnc_btree_compileTree;
```

game/functions/systems/behaviour_trees/nodes/global/fn_btree_decorator_basic.sqf
Spoffy: These are all implemented in the runner / compiler, at various points.

onTreeAssigned/Unassigned are in here:
game/functions/systems/behaviour_trees/runner/global/fn_btree_setTree.sqf
game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
game/functions/systems/behaviour_trees/runner/global/fn_btree_callOnTreeAssignedCallbacks.sqf
game/functions/systems/behaviour_trees/runner/global/fn_btree_callOnTreeUnassignedCallbacks.sqf

isService:
game/functions/systems/behaviour_trees/runner/global/fn_btree_tickGroup.sqf
game/functions/systems/behaviour_trees/runner/global/fn_btree_enterNode_decorator.sqf

abortChildrenOnConditionFailure:
game/functions/systems/behaviour_trees/runner/global/fn_btree_enterNode_decorator.sqf


They're all in use in the enemy AI tree in the later PR, except for abortChildrenOnConditionFailure. :)

game/functions/systems/behaviour_trees/compiler/global/fn_btree_compileTree.sqf
Spoffy: They're designed to run once per node type, instead of per node. 

The idea is for them to be used to perform setup/teardown to allow a node type to function, but using that node 20 times doesn't result in 20 sets of callbacks being registered. 

Something like 'setupNodeType' might even be a better name?

This should definitely be clearer in the docs though, as I think it's unintuitive. (I think a README at some point would help a lot)

game/functions/systems/behaviour_trees/compiler/global/fn_btree_runBasicTest.sqf
Spoffy: Yeah, I think you're right on this one!

game/functions/systems/behaviour_trees/nodes/global/fn_btree_decorator_basic.sqf
veteran29: What I meant is that these are not fetched from `_params` in this function and not set on the node hashmap.

Spoffy: Ah, right, I see. Yeah, assigned/unassigned are valid on any node, so probably need documenting elsewhere honestly, instead of implying they're decorator-specific.

`isService` and `abortChildrenOnConditionFailure` are added to the hashmap below for documentation reasons. Other decorator node types can use this as a base node, and overwrite them. E.g `game/functions/systems/behaviour_trees/nodes/global/fn_btree_decorator_basicService.sqf` sets `isService`, as that one should be a service node.

Spoffy: Hopefully that answers the question better! :D
If not, let me know and I'll try and clarify :)

veteran29 (approved):
LGTM

PR #119, opened 2024-02-23
Locational Events - Add system

Spoffy:
Adds a system for listening to events in specific locations, broadcast over the network (or local).

Intention is for this to be used for the AI perception system.


game/functions/systems/locational_events/global/fn_locEvents_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+    Public: No
+
+    Description:
+        Preinit for locational events.
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        The initialised events data [HashMap]
```
veteran29, line 15: With the convention we're using this function should run "by itself" via `preInit = 1` config. If it's supposed to be executed by other function it should be named just `_init`.

game/functions/systems/locational_events/global/fn_locEvents_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+private _locEventsData = localNamespace getVariable "vgm_l_locEvents_data";
+
+if (isNil "_locEventsData") then {
+    localNamespace setVariable ["vgm_l_locEvents_data", createHashMapFromArray [
+        ["listenerEventTypes", createHashMap],
+        ["perceptionGroups", createHashMap]
+    ]];
+};
+
+localNamespace getVariable "vgm_l_locEvents_data";
```
veteran29: Faster on average and easier to follow the flow IMO:
```suggestion
private _locEventsData = localNamespace getVariable "vgm_l_locEvents_data";

if (isNil "_locEventsData") then {
    _locEventsData = createHashMapFromArray [
        ["listenerEventTypes", createHashMap],
        ["perceptionGroups", createHashMap]
    ];

    localNamespace setVariable ["vgm_l_locEvents_data", _locEventsData];
};

_locEventsData // return
```

game/functions/systems/locational_events/global/fn_locEvents_removeHandlers.sqf
```diff
@@ -0,0 +1,56 @@
...
+{
+    _x params ["_type", "_id"];
+
+    private _typeListeners = _listenersByType get _type;
+    private _listenerDetails = _typeListeners get "listenerHandlers" get _listenerHash;
+
+    // If the listener has already been removed, we can skip.
+    if (isNil "_listenerDetails") then {
+        continue;
+    };
```
veteran29:
```suggestion
    if (isNil "_listenerDetails") then {continue};
```

game/functions/systems/locational_events/global/fn_locEvents_triggerEvent.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+params ["_perceptionGroup", "_pos", "_radius", "_type", "_details", ["_isLocal", false]];
+
+private _arguments = _this select [0, 5];
+
+if (_isLocal) then {
+    _arguments call vgm_g_fnc_locEvents_callHandlers;
+} else {
+    _arguments remoteExecCall ["vgm_g_fnc_locEvents_callHandlers", 0];
+};
```
veteran29:
```suggestion
if (_isLocal) exitWith {
    _arguments call vgm_g_fnc_locEvents_callHandlers;
};

_arguments remoteExecCall ["vgm_g_fnc_locEvents_callHandlers", 0];
```

game/functions/systems/locational_events/global/fn_locEvents_callHandlers.sqf
```diff
@@ -0,0 +1,49 @@
...
+    Author: Savage Game Design
+    Date: 2024-02-16
+    Last Update: 2024-02-16
+    Public: No
+
+    Description:
+        Calls handlers for a given event
+
+    Parameter(s):
+        _perceptionGroup - The perception group ID to send the event to [STRING]
```
veteran29: As mentioned in chat I think we should rename it to something like `_eventNamespace` or `_eventGroup`. In all of the functions

veteran29 (changes requested):
Some minor nitpicks ;)

game/functions/systems/locational_events/global/fn_locEvents_preInit.sqf
Spoffy: Honestly not sure what the right approach here is - I do both with this 😆 

It's preInit'd, but also called from `onNearbyEvents` just in case we end up trying to register handlers in the preInit code anywhere else.

Happy to follow whatever you reccommend here 🙂

game/functions/systems/locational_events/global/fn_locEvents_preInit.sqf
Spoffy: Wise move, I like it.

game/functions/systems/locational_events/global/fn_locEvents_callHandlers.sqf
Spoffy: Done!

game/functions/systems/locational_events/global/fn_locEvents_preInit.sqf
veteran29: It's not preInited, the function does not have the tag in the config. I would just rename it to init and call it only from `onNearbyEvents`, it's behind `isNil` check that's fired every time anyway now.

Spoffy: I'll add it to preInit, as the other functions will break unless I add the init code to all of them, which isn't something I'd like to do.

There's because there's no guarantee when `onNearbyEvents` will be called for the first time, and the rest of the functions assume the events data already exists.

veteran29 (approved):
LGTM

PR #120, opened 2024-03-01
Director - Broadcast mission groups creation to participating players

veteran29:
- title
- JIP handling for client-side monitoring (from TODO)

### DIFF
--- a/game/functions/systems/mission_director/server/fn_director_startMission.sqf
+++ b/game/functions/systems/mission_director/server/fn_director_startMission.sqf
@@ -2,7 +2,7 @@
     File: fn_director_startMission.sqf
     Author: Savage Game Design
     Date: 2023-09-23
-    Last Update: 2023-11-04
+    Last Update: 2024-03-01
     Public: Yes
 
     Description:
@@ -55,8 +55,17 @@ private _jobId = format ["missionDirector%1", _mission get "public" get "id"];
 
 _directorData set ["schedulerJob", _jobId];
 
-// TODO - Start clientside monitoring on JIP players
-// Use startDeploy and finishDeploy as events for ease of use.
+// Handle JIPs
+private _attachedEh = ["vgm_mission_attached", [_mission, {
+    params ["_args", "_mission"];
+    _args params ["_playerId", "_missionId"];
 
+    if ((_mission get "public" get "id") != _missionId) exitWith {};
+    getUserInfo _playerId params ["", "_machineId"];
 
+    [] remoteExec ["vgm_c_fnc_director_startClientsideMonitoring", _machineId];
 
+    ["vgm_mission_director_squadCreated", _mission get "director" get "aiGroups", _machineId] call para_g_fnc_event_triggerTargets;
+}]] call para_g_fnc_event_subscribeLocal;
+
+_directorData set ["eventHandlers", [_attachedEh]];
--- a/game/functions/systems/mission_director/server/fn_director_startMission.sqf
+++ b/game/functions/systems/mission_director/server/fn_director_startMission.sqf
@@ -54,18 +54,3 @@ private _jobId = format ["missionDirector%1", _mission get "public" get "id"];
 [_jobId, { _this call vgm_s_fnc_director_processMission }, [_mission], 10] call para_g_fnc_scheduler_add_job;
 
 _directorData set ["schedulerJob", _jobId];
-
-// Handle JIPs
-private _attachedEh = ["vgm_mission_attached", [_mission, {
-    params ["_args", "_mission"];
-    _args params ["_playerId", "_missionId"];
-
-    if ((_mission get "public" get "id") != _missionId) exitWith {};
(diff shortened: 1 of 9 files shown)


game/functions/systems/mission_director/server/fn_director_startMission.sqf
```diff
@@ -55,8 +55,17 @@ private _jobId = format ["missionDirector%1", _mission get "public" get "id"];
 
 _directorData set ["schedulerJob", _jobId];
 
-// TODO - Start clientside monitoring on JIP players
-// Use startDeploy and finishDeploy as events for ease of use.
+// Handle JIPs
+private _attachedEh = ["vgm_mission_attached", [_mission, {
+    params ["_args", "_mission"];
+    _args params ["_playerId", "_missionId"];
```
Spoffy: Is this getting the correct args?

game/functions/systems/mission_director/server/fn_director_startMission.sqf
```diff
@@ -55,8 +55,17 @@ private _jobId = format ["missionDirector%1", _mission get "public" get "id"];
...
 
+    if ((_mission get "public" get "id") != _missionId) exitWith {};
+    getUserInfo _playerId params ["", "_machineId"];
 
+    [] remoteExec ["vgm_c_fnc_director_startClientsideMonitoring", _machineId];
 
+    ["vgm_mission_director_squadCreated", _mission get "director" get "aiGroups", _machineId] call para_g_fnc_event_triggerTargets;
```
Spoffy: At some point, I think I'd like this to be closer to the other place the event is emitted. Having it in a separate handler + file I think obscures the behaviour of `squadCreated` (i.e, when the event is triggered).

Right now, I'm wondering if something like:
```
[
   _codeToCallNow,
   _codeToCallOnJIP,
   _jipEventID
] call vgm_s_fnc_mission_triggerEvent
```
Would be a useful function to have. 
Or
```
[
   _jipID,
   _code
] call vgm_s_fnc_mission_onJip
```
...But I'm not sure yet. I think it's probably too early to refactor into the right patterns, so I'm inclined to leave it for now? :)

game/functions/systems/mission_director/server/fn_director_startMission.sqf
```diff
@@ -55,8 +55,17 @@ private _jobId = format ["missionDirector%1", _mission get "public" get "id"];
...
+}]] call para_g_fnc_event_subscribeLocal;
```
Spoffy: Could we register this in `preInit`? It feels like we don't need to register this on a per-mission basis, since we've got access to the director data through `_mission get "director"`. 

What do you think?

Spoffy (changes requested):

game/functions/systems/mission_director/server/fn_director_startMission.sqf
Spoffy: Thinking about this more - I don't think we should send `squadCreated`  on JIP,  as the events are occurring in different contexts.

I think I'd prefer a separate event, `initialDirectorData` or `initialSquadData`  or `jipSquadData`, to distinguish the JIP case from the "I'm already on a mission and a new squad spawns" situation. It gives us more power on the client to do extra things for JIP.

game/functions/systems/mission_director/server/fn_director_startMission.sqf
veteran29: I think so? I've tested it by JIP attaching the player via console and it worked properly.

Event is fired with:
```sqf
["vgm_mission_attached", [_playerId, _missionPublic get "id"]] call para_g_fnc_event_triggerGlobal;
```
Event system does:
```sqf
[_this param [DATA_PARAMETER_INDEX], _savedParameters, _eventName, _topic, _originMachineId] call _code;
```

game/functions/systems/mission_director/server/fn_director_startMission.sqf
veteran29: Something like `vgm_s_fnc_mission_onJip` would work, we can easily refactor this if needed.

veteran29: Changed the JIP event to `vgm_mission_director_jipData` to have it separate.

game/functions/systems/mission_director/server/fn_director_startMission.sqf
Spoffy: Re-read this today, and realised I'm just being an idiot. I mixed up the event `_args` with the saved parameters that get passed to the function (`_mission`). 

I usually use `_eventData` instead of `_args`, which I think is where the confusion came from :D

game/functions/systems/mission_director/server/fn_director_startMission.sqf
Spoffy: Perfect, thanks :D

Spoffy (approved):

PR #121, opened 2024-03-01
Skill Investigate - Tweaks

veteran29:
- Increase waves distance to 200m
- Show tiny waves for close static enemies
- Show sound waves for units speaking via SAM system
- Show sound waves for unit shots

Needs #120

### DIFF
--- a/game/functions/functions_client.hpp
+++ b/game/functions/functions_client.hpp
@@ -518,6 +518,7 @@ class vgm_c
         VGM_CLIENT_PATH(\systems\skill_investigate\client);
 
         class skill_investigate_addAction {};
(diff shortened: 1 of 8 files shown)


Spoffy (approved):
Nice!

My only question is whether we could get a performance boost, from having Fired / Voices event handlers only run when the player is in listen mode.

Either by adding/removing the handlers, or a straightforward `if (isNil "vgm_c_skill_investigate_noises")`?

veteran29:
I think the boost would be negligible in this case. At some point, I will want to record the whole session in the profiler and look for potential optimization areas.

PR #124, opened 2024-03-02
Nightvision - Buff the effect for strong overcast

veteran29:
- title

The second stronger version of set aperture was never applied as its condition was inclusive in the first condition so it never was executed. (also present in current version of MF/Paradigm looking at the source)

### DIFF
--- a/game/functions/core/client/ui/fn_handle_light_level_loop.sqf
+++ b/game/functions/core/client/ui/fn_handle_light_level_loop.sqf
@@ -2,26 +2,27 @@
     File: fn_handle_light_level_loop.sqf
     Author: Gus Schultz
     Date: 2022-09-18
-    Last Update:
-    Public: /shrug
+    Last Update: 2024-03-02
+    Public: No
 
     Description:
-        used to set camera aperature based on light level
+        Used to set camera aperature based on light level
 
     Parameter(s):
-        N/A
+        None
 
-    Returns: nothing
+    Returns:
+        Nothing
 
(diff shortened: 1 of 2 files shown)


veteran29:
Might be too strong now, however the exampe is near the shore, it's darker in the jungle.

```sqf
setApertureNew [4, 6, 9, 0.9];
```
image

```sqf
private _minAperture = linearConversion [0, 4, _lightBrightness, 1, 3, true];
setApertureNew [_minAperture, 6, 9, 0.9];
```
image

Spoffy (changes requested):
Is this definitely the fix? The comparison doesn't look right to me now.

What I see is this:
```
if (_lightBrightness < 4 && _lightBrightness < 120) exitWith { ... }

if (_lightBrightness <= 4 exitWith { ... }
```

veteran29:
You're completly right! I hate conditions like this where the variable is not always on the same side, my brain completely ignored that.

Fixed this and added slight buff for the nightvision on higher overcast, which I think was the issue on our playtests.

Spoffy (approved):

Spoffy:
> You're completly right! I hate conditions like this where the variable is not always on the same side, my brain completely ignored that.

Makes total sense! I stared at it for a good few minutes making sure I wasn't missing something :sweat_smile: 

The reason I do it that way is because it's similar to this: `4 < x < 120`, i.e the 'Between 4 and 120' expression. 
That said, happy for it to be changed :D

PR #125, opened 2024-03-02
WIP: Para Cleanup - Do not delete bodies close to players

veteran29:
- title

Bodies near the players will be skipped, searches until it finds the bodies to clear or the array ended.

### DIFF
--- a/game/mission/para_server_init.sqf
+++ b/game/mission/para_server_init.sqf
@@ -60,6 +60,7 @@ diag_log "VGM: Initialising Cleanup Routine";
 [
     createHashmapFromArray [
         ["minPlayerDistance", ["cleanup_min_player_distance", 400] call BIS_fnc_getParamValue],
+        ["minPlayerBodyDistance", ["cleanup_min_player_body_distance", 150] call BIS_fnc_getParamValue],
         ["maxBodies", ["cleanup_max_bodies", 50] call BIS_fnc_getParamValue],
         ["cleanPlacedGear", ["cleanup_placed_gear", 1] call BIS_fnc_getParamValue > 0],
         ["placedGearCleanupTime", ["cleanup_placed_gear_lifetime", 300] call BIS_fnc_getParamValue],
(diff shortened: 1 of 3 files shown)


Spoffy (changes requested):
I'm not sure this is the right way to solve this!

The reason behind adding overrunning bodies to `para_s_cleanup_items_delete_immediately` is performance. The "maximum 50 bodies" is a *hard* cap. 

If it becomes a soft cap, controlled by proximity to the player, the players could get swarmed, and surrounded by 500 bodies if they remain still. That's the situation that happened in MF, where players are often fighting in a relatively small area for long periods.

## Suggested fix
My suggested fix for this is actually to enable Arma 3's native garbage collection for bodies (which is enabled in Mike Force), and allow this to keep running as intended as an emergency fallback.

That means the oldest bodies will get cleaned up first, so players will see far fewer nearby bodies disappearing. 

## Alternative fix
The *alternative* would be to add dead bodies to this cleanup system on a time limit, by tweaking the "EntityKilled" handler in "fn_cleanup_subsystem_init", but I don't think we gain anything from having this system handle time-based deletion of bodies when Arma does those natively?

## Proposed changes either way
We could potentially increase the max dead bodies a little too, if we feel the need for more wiggle-room. Particularly given we'll have multiple teams on the mission, 100 feels more sane.

We could also make the emergency hard cap delete from the *front* of the bodies array, meaning we delete the oldest instead of the newest bodies.

veteran29:
> My suggested fix for this is actually to enable Arma 3's native garbage collection for bodies

the Arma garbage collector is riddled with bugs and is unreliable, KK is working on fixing it but it most likely will be fixed in 2.18 (2.16 is next update) - https://feedback.bistudio.com/T171981

My issue with overrun handling is bodies vanishing right when you want to loot them which is more likely the more groups are playing on the server. This is pretty jarring and bad from gameplay flow.

Spoffy:
> the Arma garbage collector is riddled with bugs and is unreliable, KK is working on fixing it but it most likely will be fixed in 2.18 (2.16 is next update) - https://feedback.bistudio.com/T171981

In that case, I'd propose we flip the body deletion to do the start of the array (oldest body first), and bump the limit up to 75 or 100.

That should probably solve 90% of the situations, without incurring a performance hit I'd think :D

veteran29:
Can reopen when needed.

PR #131, opened 2024-04-04
Missions selection - Add zone reservation and modifiers system

veteran29:
- title

### DIFF
--- a/game/functions/systems/missions_zones/client/remoteExec/fn_missions_zones_remoteExec_receiveList.sqf
+++ b/game/functions/systems/missions_zones/client/remoteExec/fn_missions_zones_remoteExec_receiveList.sqf
@@ -1,0 +1,31 @@
+/*
+    File: fn_missions_zones_remoteExec_receiveList.sqf
+    Author: Savage Game Design
+    Date: 2024-04-04
+    Last Update: 2024-04-04
+    Public: No
+
+    Description:
+        Receive missions zone data and open missions display.
+
+    Parameter(s):
+        _timeToMissionsChange - Time after which new missions will be available [NUMBER]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        [_seedChangeIn, vgm_missions_zones_targetBoxesModifiers] remoteExecCall ["vgm_c_fnc_missions_zones_remoteExec_receiveList", remoteExecutedOwner];
+ */
+
+params ["_timeToMissionsChange", "_targetBoxesModifiers"];
+
+// player aborted while display was loading
+if (isNull (uiNamespace getVariable ["RscDisplayCommonMessage_display", displayNull])) exitWith {};
+
+// close the popup
+private _result = true;
+uiNamespace setVariable ["BIS_fnc_guiMessage_status", _result];
+
+// TODO mission display implementation
+hint str _this;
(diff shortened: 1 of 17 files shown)


game/functions/systems/missions_zones/server/fn_missions_zones_freeZone.sqf
```diff
@@ -0,0 +1,29 @@
+/*
+    File: fn_missions_selection_freeZone.sqf
```
Spoffy: Old folder name

game/functions/systems/missions_zones/client/remoteExec/fn_missions_zones_remoteExec_receiveList.sqf
Spoffy, line 1: With this "fetch list" and "receive list" approach - I wonder if we should just keep the client updated with zone info at all times via a netmap (or similar). It won't change often, and might make the logic here simpler?

game/functions/systems/missions_zones/server/fn_missions_zones_reserveZone.sqf
```diff
@@ -0,0 +1,34 @@
...
+private _zoneReservationNetmap = ["vgm_missions_zones_zoneReservations"] call para_g_fnc_netmap_get;
+
+if (_targetZone in _zoneReservationNetmap) exitWith {
+    format ["Unable to reserve zone %1 for %2", _targetZone, _playerId] call vgm_g_fnc_logInfo;
+    false
+};
+
+format ["Reserved zone %1 for %2", _targetZone, _playerId] call vgm_g_fnc_logInfo;
+
+[_zoneReservationNetmap, _targetZone, _playerId] call para_s_fnc_netmap_set;
```
Spoffy, line 33: I wonder if zones should be reserved by a group, in case the mission leader quits or is disconnected?

If we do that, we'll also need to free up the reservation if the group is empty or deleted.

Spoffy: Alternatively we could have a "isZoneReserved" function that does all these checks? Might be simpler than handling the edge cases with all the additional event handlers.

Spoffy (changes requested):

game/functions/systems/missions_zones/server/fn_missions_zones_reserveZone.sqf
veteran29: I think it's fine as is for now, player ID does change if you do a full reconnect, leader leaving is not an issue as the zone is auto freed when mission is terminated.

game/functions/systems/missions_zones/client/remoteExec/fn_missions_zones_remoteExec_receiveList.sqf
veteran29: In general, I agree but in this particular case it's using these functions to prompt the eventual regeneration of the zone modifiers in the [getList](https://github.com/Savage-Game-Design/Across-the-Fence/pull/131/files#diff-b661f09cfdfccd3964fab0851b84a69c72039343ec4a5d9aa4ae27aa1b5aea81R38) function. 

I think this should be revised once we'll start implementing the "alive" zones that change after every mission.

game/functions/systems/missions_zones/server/fn_missions_zones_reserveZone.sqf
Spoffy: Alright, let's merge for now, and we'll revisit when we get better mission/group management mechanics!

PR #133, opened 2024-05-11
Rewrites Paradigm's keyhandler

Spoffy:
This PR:
* Rewrites paradigm's keyhandler system, and expands it to work correctly when displays are opened by the gamemode.
* Converts the skill system to use keyhandler.
* Adds the keybinding menu to VGM


paradigm/client/functions/keyhandler/fn_keyhandler_registerAction.sqf
```diff
@@ -0,0 +1,37 @@
+/*
+    File: fn_keyhandler_registerAction.sqf
+    Author: Savage Game Design
+    Date: 2024-05-11
+    Last Update: 2024-05-11
+    Public: No
```
veteran29: This looks like it should be used by other modules.
```suggestion
    Public: Yes
```

paradigm/client/functions/keyhandler/fn_keyhandler_registerAction.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+    Returns:
+        Nothing
+
+    Example(s):
+        // See para_c_fnc_keyhandler_init for structure of _myActionHashMap
+        [_myActionHashMap] call para_c_fnc_keyhandler_registerAction;
+ */
+
+params ["_action"];
```
veteran29: If this is supposed to be a public API I would add type check to params.
```suggestion
params [["_action", createHashmap]];
```

game/functions/systems/medical/client/fn_medical_preInit.sqf
```diff
@@ -119,3 +121,16 @@ vgm_c_medical_damageModifiers = [];
         true // stop processing
     };
 }] call vgm_c_fnc_medical_addDamageModifier;
+
+[
+    createHashMapFromArray [
+        ["name", "OpenMedicalMenu"],
```
veteran29, line 127: Name is used as a key in profile namespace save data. IMO should be prefixed.
```suggestion
        ["name", "vgm_OpenMedicalMenu"],
```

game/functions/systems/medical/client/fn_medical_preInit.sqf
```diff
@@ -119,3 +121,16 @@ vgm_c_medical_damageModifiers = [];
...
+        ["displayName", "STR_VGM_MEDICAL_UI_OPEN_MEDICAL_MENU"],
+        ["onRelease", false],
+        ["defaultKey", createHashMapFromArray [
+            ["dikCode", DIK_H]
+        ]]
+    ]
+] call para_c_fnc_keyhandler_registerAction;
+
+["OpenMedicalMenu", vgm_c_fnc_medical_openMedicalMenu] call para_c_fnc_keyhandler_addGeneralActionHandler;
```
veteran29, line 136: Goes with above suggestion.
```suggestion
["vgm_OpenMedicalMenu", vgm_c_fnc_medical_openMedicalMenu] call para_c_fnc_keyhandler_addGeneralActionHandler;
```

paradigm/client/functions/keyhandler/fn_keyhandler_addDisplayActionHandler.sqf
```diff
@@ -0,0 +1,31 @@
...
+    Parameter(s):
+        _display - Display to register the handler to [DISPLAY]
+        _actionName - Action to register for [STRING]
+        _handler - Code to fire [CODE]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        ["MyAction", para_c_fnc_doThing] call para_c_fnc_keyhandler_addGeneralActionHandler;
```
veteran29: Invalid example

paradigm/client/functions/keyhandler/fn_keyhandler_enableGeneralActionTriggeringOnDisplay.sqf
```diff
@@ -0,0 +1,27 @@
...
+    Parameter(s):
+        _display - Display to enable this feature on [DISPLAY]
+
+    Returns:
+        Nothing
+
+    Example(s):
+        private _display = createDisplay _myDisplay;
+        [_display] call para_c_fnc_keyhandler_enableGeneralActionTriggeringOnDisplay;
+ */
```
veteran29, line 22: Not feeling the purpose of this function, it only forwards the call to `hookDisplayKeyEvents`.

paradigm/client/functions/keyhandler/fn_keyhandler_getAllKeybinds.sqf
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+private _registeredActions = localNamespace getVariable "para_keyhandler_actions";
+private _result = createHashMap;
+
+{
+    if (_x in _registeredActions) then {
+        _result set [_x, _y];
+    };
+} forEach (profileNamespace getVariable "para_c_keyhandler_bindings");
```
veteran29: Is the storage format different for new system than in the old one? If so I would rename the `para_c_keyhandler_bindings` or add v2 suffix so it does not break if people play this and mikeforce.

veteran29 (changes requested):

paradigm/client/functions/keyhandler/fn_keyhandler_enableGeneralActionTriggeringOnDisplay.sqf
Spoffy: The main purpose of this function is to avoid having to remember how to do this when working on new displays in the future - even if "this" is just enabling a variable and making the display aware of key events.

As side benefits it gives us:
- 1 function to search for when looking for "global keybind aware" displays.
- A single point to refactor if we need to change this behaviour in the future.

paradigm/client/functions/keyhandler/fn_keyhandler_getAllKeybinds.sqf
Spoffy: Didn't even realise I'd used the same variable, amazing catch!

game/functions/systems/medical/client/fn_medical_preInit.sqf
Spoffy: Mitigated this by prefixing the profile namespace key instead on a per-gamemode basis, so each gamemode has its own hashmap.

paradigm/client/functions/keyhandler/fn_keyhandler_init.sqf
```diff
@@ -0,0 +1,87 @@
...
+            ["ctrl", true],
+            ["alt", true]
+        ]]
+    ]
+]
+*/
+
+params ["_gamemodeId"];
+
+para_c_keyhandler_gamemodeId = _gamemodeId;
```
veteran29, line 67: I would add an error/warning here in case it's double initialized:
```suggestion
if (!isNil "para_c_keyhandler_gamemodeId") then {
    ["ERROR", format ["Keybinding system was already initialized: %1", _gamemodeId]] call para_g_fnc_log;
};

para_c_keyhandler_gamemodeId = _gamemodeId;
```

veteran29 (approved):

PR #156, opened 2024-08-21
Indexes LZ locations

Spoffy:
Adds LZ locations to the terrain indexes.

Decouples the location system from the site system, to allow us to inject custom location types without needing new sites.

(Using sites would complicate all future spawning)


veteran29 (approved):
LGTM

PR #160, opened 2024-08-23
Missions - Zoom map on AO and cover map during mission start

veteran29:
- title

Cover map is copy pasted from BI module so don't review it too hard. ;)

image

### DIFF
--- a/game/functions/systems/missions/client/internal/fn_missions_finishDeploy.sqf
+++ b/game/functions/systems/missions/client/internal/fn_missions_finishDeploy.sqf
@@ -2,7 +2,7 @@
     File: fn_missions_finishDeploy.sqf
     Author: Savage Game Design
     Date: 2023-02-26
-    Last Update: 2024-08-23
+    Last Update: 2024-08-24
     Public: No
 
     Description:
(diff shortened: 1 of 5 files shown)


game/functions/systems/missions/client/internal/fn_missions_finishDeploy.sqf
```diff
@@ -43,6 +43,17 @@ _this spawn {
 
     ["vgm_mission_deploy_local", _currentMission] call para_g_fnc_event_triggerLocal;
 
+    // zoom map on the mission area
+    [] call vgm_c_fnc_missions_coverMap;
+    vgm_missions_zoomOnMapScript = (_currentMission get "targetZone") spawn {
+        waitUntil {visibleMap}; // can't animate hidden map
+        [
+            markerSize _this vectorMultiply 2.5,
+            markerPos _this,
```
Spoffy: This will break with the updated target box implementation.

`format ["tbox_%1", _this]` should work as a marker name.

Spoffy (changes requested):

game/functions/systems/missions/client/internal/fn_missions_finishDeploy.sqf
Spoffy: Fixing this and pushing

PR #161, opened 2024-08-24
Scouting - Add scouting system

veteran29:
- title


game/functions/systems/sites/server/fn_sites_spawn.sqf
```diff
@@ -31,9 +31,14 @@ if (isNil "_siteType") exitWith {
...
+private _site = createHashMapFromArray [
+    ["id", hashValue _pos2D],
     ["type", _siteType],
     ["pos", _pos2D],
     ["objects", _spawnResult # 0],
     ["ownedSites", []]
-]
+];
+
+["vgm_sites_siteSpawned", [_site]] call para_g_fnc_event_triggerServer;
```
Spoffy: Since this is a server function, we could use `triggerLocal` which would be more efficient, as it avoids the remoteExec call.

veteran29:
After the playtest will be reworked for manual site type and grid input.

game/functions/core/client/displays/fn_displayNotepad.sqf
```diff
@@ -0,0 +1,473 @@
...
+            _display setVariable ["vgm_onMapId", _ehId];
+        };
+
+        // debug background
+        if (false) then {
+            private _ctrlBg = _display ctrlCreate ["VGM_ctrlStatic", -1, _ctrlGrpMain];
+            _ctrlBg ctrlSetBackgroundColor [1,0,0,0.35];
+            _ctrlBg ctrlSetPosition [0, 0, VGM_NOTEPAD_W, VGM_NOTEPAD_H];
+            _ctrlBg ctrlCommit 0;
+        };
```
Spoffy: Do we want to hide this behind a __A3_DEBUG__ or some setting that can be enabled at runtime? Otherwise can we remove it or comment it out?

game/functions/core/client/displays/fn_displayNotepad.sqf
```diff
@@ -0,0 +1,473 @@
...
+
+        private _ctrlHeader = _display ctrlCreate ["VGM_ctrlStaticNotepadHeader", -1, _ctrlGrpMain];
+        _ctrlHeader ctrlSetFontHeight (VGM_NOTEPAD_LINE_H * 1.8);
+        _ctrlHeader ctrlSetText toUpper "Scouting report";
```
Spoffy: String needing localizing?

game/functions/systems/sites/server/fn_sites_spawn.sqf
```diff
@@ -47,9 +47,15 @@ if (_siteType getOrDefault ["hideNearbyTerrain", false]) then {
 
 private _spawnResult = [_pos2D] call (_siteType get "spawnFunction");
 
-createHashMapFromArray [
+private _site = createHashMapFromArray [
+    ["id", hashValue _pos2D],
+    ["class", _typeId],
```
Spoffy, line 42: I feel like "class" might be a bit broad of a name. We could just use "icon" perhaps? Depends how you're thinking we should use this in the future.

game/functions/systems/missions/server/fn_missions_createSystemNetmap.sqf
```diff
@@ -0,0 +1,44 @@
...
+    Author: Savage Game Design
+    Date: 2024-08-13
+    Last Update: 2024-08-24
+    Public: Yes
+
+    Description:
+        Create system netmap in specified mission id.
+
+    Parameter(s):
+        _missionId - Id of the missin to get the netmap from [NUMBER]
```
Spoffy:
```suggestion
        _missionId - Id of the mission to get the netmap from [NUMBER]
```

game/functions/systems/missions/server/fn_missions_getSystemNetmap.sqf
Spoffy, line 1: Should we make this a global function? Since we're using netmaps, it feels natural to make the mission's system netmaps available on the client via a nice function call.

Can use vn_c_fnc_getMissions for that (which actually should be global itself, since it will happily work globally)

game/functions/systems/missions/server/fn_missions_startMission.sqf
```diff
@@ -37,11 +37,14 @@ if (_missionPublic get "status" isNotEqualTo "CREATED") exitWith {
 
 [] remoteExecCall ["vgm_c_fnc_missions_startDeploy", values (_mission get "machineIds")];
 
-[_missionPublic get "targetZone", 25] call vgm_s_fnc_missions_zones_spawnRandomSites;
+private _sites = [_missionPublic get "targetZone", 25] call vgm_s_fnc_missions_zones_spawnRandomSites;
+(_mission get "sites") append _sites;
```
Spoffy: Rather than tracking this with the `sites` variable within the mission system, you can use vgm_s_fnc_missions_zones_getSites to fetch the sites.

That's because sites are more closely connected to a zone rather than a specific mission (they'll be semi-persistent in zones eventually).

game/functions/systems/missions/server/fn_missions_createMission.sqf
```diff
@@ -53,7 +53,9 @@ private _mission = createHashMapFromArray [
     // Copy the parameters hashmap to prevent it being accidentally modified elsewhere.
     ["parameters", +_parameters],
     // Maps player ID to their client ID for easy remoteExec'ing
-    ["machineIds", createHashMap]
+    ["machineIds", createHashMap],
+    // Other
+    ["sites", []]
```
Spoffy: This can probably be removed, and `vgm_s_fnc_missions_zones_getSites` used instead (see a later comment)

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
```diff
@@ -29,4 +29,43 @@ if (_endType == "SUCCESS") then {
     _milestones pushBack ["mission_success", 375];
 };
 
+// add XP for spotting
+call {
+    private _mission = [_playerId] call vgm_s_fnc_missions_getAssignedMission;
+    private _scoutingData = [_mission get "public" get "id", "scouting"] call vgm_s_fnc_missions_getSystemNetmap;
+
+    private _missionSites = +(_mission get "sites");
```
Spoffy:
```suggestion
    private _missionSites = +((_mission get "public" get "targetZone") call vgm_s_fnc_missions_zones_getSites);
```

Can do something like this, and not track `sites` inside the mission.

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
```diff
@@ -29,4 +29,43 @@ if (_endType == "SUCCESS") then {
...
+    private _guessedSites = _scoutingData get "guessedSites";
+    {
+        #define SITE_DIST ((_x get "pos") distance2d _guessedPos)
+
+        _x params ["", "_guessedClass", "", "_guessedPos", "_guessId"];
+        private _guessIdText = parseNumber _guessId + 1;
+
+        private _nearSites = _missionSites select {SITE_DIST <= 200};
```
Spoffy: Might be worth switching to using `inAreaArrayIndexes` here, to prevent any performance hiccups in the future.

This select is currently quadratic in the number of sites, so 500 in the worst case right now (number_of_guessed_sites * number_of_sites), and number of guessed sites is capped at the number of sites.

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
```diff
@@ -29,4 +29,43 @@ if (_endType == "SUCCESS") then {
...
+        } else {
+            _milestones pushBack ["site_spotted_good", 100, [_guessIdText]];
+
+            _nearSites apply {[SITE_DIST, _x]};
+            _nearSites sort false;
+
+            private _goodMatch = _nearSites select 0;
+
+            _missionSites deleteAt (_missionSites find _goodMatch);
+        };
```
Spoffy: I read this as:

```
If there is a site of the right type within 200m:
    Score 200, prevent this site from scoring again.
    
Else if there is a site of the wrong type within 200m:
    Score 100, prevent this site from scoring again.
```

One issue here is it means a good spot can prevent a perfect spot - if you spot a nearby site with the wrong type, and also spot correctly the site with the right type, there's a chance the good spot gets processed first.

I think that's okay for now - but could we add a comment explaining it, so we don't get caught out in the future?

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_postInit.sqf
```diff
@@ -0,0 +1,58 @@
...
+
+vgm_scouting_locations = createHashMap;
+
+vgm_scouting_siteTypes = "getNumber (_x >> 'disabled') == 0" configClasses (missionConfigFile >> "vgm_site_types") apply {
+    [
+        localize getText (_x >> "displayNameKey"), // name first, for sorting
+        configName _x,
+        getText (_x >> "locationClass")
+    ]
+};
```
Spoffy, line 22: An alternative here is to use `vgm_s_fnc_sites_getAllSiteTypes`, instead of reading from the config. They're initialised in preInit, so it would be fine to use here :)

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_postInit.sqf
```diff
@@ -0,0 +1,58 @@
...
+    (_this#0) params ["_siteId", "_player"];
+
+    ["VGM_SiteTypeChanged", [name _player, parseNumber _siteId + 1]] call BIS_fnc_showNotification;
+    [_siteId] call vgm_c_fnc_missions_gameplay_scouting_createUpdateLocation;
+
+}] call para_g_fnc_event_subscribeServer;
+
+["vgm_mission_started", {
+    {deleteLocation (vgm_scouting_locations deleteAt _x)} forEach vgm_scouting_locations;
+}] call para_g_fnc_event_subscribeServer;
```
Spoffy, line 58: I love how cleanly this works with the event-driven approach.

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
```diff
@@ -0,0 +1,41 @@
...
+    Public: No
+
+    Description:
+        No description added yet.
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
```
Spoffy: Documentation 🙂

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_getSiteById.sqf
Spoffy, line 1: We could make this global, if we made a couple of other netmap-related mission functions global (e,g "getSystemNetmap").

That would solve your "TODO - Make shared function" comments in most of the server's handlers for finding the guessed site.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_handleSpotted.sqf
```diff
@@ -0,0 +1,52 @@
...
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        [player, cursorObject] call vgm_s_fnc_missions_gameplay_scouting_handleSpotted;
+ */
```
Spoffy: Documentation 🙂

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_postInit.sqf
```diff
@@ -0,0 +1,62 @@
...
+    } forEach (_site get "objects");
+}] call para_g_fnc_event_subscribeServer;
+
+// user input related
+call {
+    ["vgm_scouting_spottedTarget", {
+        (_this#0) params ["_spotter", "_target"];
+        [_spotter, _target] call vgm_s_fnc_missions_gameplay_scouting_handleSpotted;
+    }] call para_g_fnc_event_subscribe;
+
```
Spoffy: Is this used? Couldn't find any code that emits this event.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_handleSpotted.sqf
```diff
@@ -0,0 +1,52 @@
+/*
+    File: fn__missions_gameplay_scouting_handleSpotted.sqf
```
Spoffy: Is this still used? The event that calls this never seems to be triggered anywhere.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_onMissionStarted.sqf
```diff
@@ -0,0 +1,66 @@
...
+
+    Parameter(s):
+        _missionId -
+
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        [_missionId] call vgm_s_fnc_missions_gameplay_scouting_onMissionStarted
+ */
```
Spoffy: Docs :)

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_postInit.sqf
```diff
@@ -0,0 +1,62 @@
...
+["vgm_sites_siteSpawned", {
+    (_this#0) params ["_site"];
+
+    {
+        if (_x isKindOf "Land_vn_vegetation_base") then {continue};
+        [_x, _site] call vgm_s_fnc_missions_gameplay_scouting_setSpottable;
+    } forEach (_site get "objects");
+}] call para_g_fnc_event_subscribeServer;
+
```
Spoffy, line 32: Might want to monitor the performance of this - there could be a *lot* of objects here. There's also potentially other implementations of taking photos we can look into (such as raycasting + using the hit position, and comparing it to the site location. The locational events system would make this simple - register each site as a listener, and have a player taking a photo trigger a "photo taken" event at the raycast hit position with a radius of 50m or similar.)

game/functions/systems/sites/server/fn_sites_spawn.sqf
Spoffy: We'll also need to move this onto the client once sites have local objects, or similar. That's going to be its own set of headaches - since you're reading the `objects` array, we'll need to figure out a way to add the local-only objects too.

game/mission/stringtable.xml
```diff
@@ -193,6 +193,15 @@
...
+        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_PERFECT">
+            <Original>%1. Perfect spot</Original>
+        </Key>
+        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_GOOD">
+            <Original>%1. Good spot</Original>
+        </Key>
+        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_BAD">
+            <Original>%1. Invalid spot</Original>
+        </Key>
     </Package>
```
Spoffy:
```suggestion
        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_PERFECT">
            <Original>%1. Enemy site marked [Perfect]</Original>
        </Key>
        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_GOOD">
            <Original>%1. Enemy site marked [Good]</Original>
        </Key>
        <Key ID="STR_VGM_MISSION_END_UI_MILESTONE_SITE_SPOTTED_BAD">
            <Original>%1. No enemy site at marked location.</Original>
        </Key>
    </Package>
```

Spoffy (changes requested):

game/functions/systems/sites/server/fn_sites_spawn.sqf
veteran29: Local objects can be stored separately as spawn ids, local objects framework supports init code which executes where it's spawned. So I don't think that'll be much of a problem. Minor changes might be needed.

game/functions/systems/sites/server/fn_sites_spawn.sqf
veteran29: I don't think `icon` is good name either. the `_typeId` is literally mapped to class in the config so I think it's fitting.

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_postInit.sqf
veteran29: This is used on the client so that data is unavailable, I think unifying this should be a part of general refactor of how sites are stored.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_postInit.sqf
veteran29: The `setSpottable` function is pretty light so it shouldn't be too hard on performance and this happens only once upon mission start so should be ok.

game/functions/systems/missions/server/fn_missions_getSystemNetmap.sqf
veteran29: I guess we can refactor when needed? Or make client variant of the function?

game/functions/core/client/displays/fn_displayNotepad.sqf
veteran29: Commented out

game/functions/core/client/displays/fn_displayNotepad.sqf
veteran29: Done

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
veteran29: `inAreaArrayIndexes` would indeed be fastest but I need to check the element `class` to check if it's a perfect spot. `inAreaArray` will still be faster than the select so I'll switch to that.

```sqf
sites = [];
sites resize 10000;
sites = sites apply {createHashMapFromArray [["pos", [random 1000, random 1000]]]};

_fnc_a = {
	_this select {(_x get "pos") distance [100, 100] < 200};
};

_fnc_b = {
	(_this apply {_x get "pos"}) inAreaArrayIndexes [[100, 100], 200, 200];
};

_fnc_c = {
	(_this apply {_x get "pos"}) inAreaArray [[100, 100], 200, 200];
};

[
	diag_codePerformance [_fnc_a, sites, 1000, true], // 5.45592 ms
	diag_codePerformance [_fnc_b, sites, 1000, true], // 1.04804 ms
	diag_codePerformance [_fnc_c, sites, 1000, true]  // 2.64881 ms
]
```

game/functions/systems/missions/server/fn_missions_getSystemNetmap.sqf
Spoffy: Yeah, sounds good - happy to refactor when needed.

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
Spoffy: Perfect, thanks for interesting performance comparison too :D

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_postInit.sqf
Spoffy: Sounds good to me 👍

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_postInit.sqf
Spoffy: Righto - we'll monitor this then! :)

game/functions/systems/missions/server/internal/fn_missions_calculateMilestones.sqf
veteran29: Used `inAreaArrayIndexes` in the end, was a bit tired and haven't noticed that `inAreaArray` will give me the same troubles so went with minor refactor and used the faster method.

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_getSiteById.sqf
veteran29: I would like to tackle this in separate PR after hints will be done.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_handleSpotted.sqf
veteran29: Removed the function, unused.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_handleSpotted.sqf
veteran29: It's a remnant from the time where scouting checked if you have "hit" the site while placing it on the map. Removed.

game/functions/systems/missions_gameplay/server/scouting/fn_missions_gameplay_scouting_postInit.sqf
veteran29: Not used, removed.

Spoffy (approved):

PR #184, opened 2024-11-17
Tutorial - Add basic framework

Spoffy:
Adds support for CfgHints entries to show as a survival card hint, when triggered by one of these conditions:
- Entering a specific marker on the map
- The client receiving a specific event from any source
- Looking at an object with `vgm_c_tutorial` set to the path of the tutorial.

### DIFF
--- a/game/functions/systems/tutorial/client/fn_tutorial_postInit.sqf
+++ b/game/functions/systems/tutorial/client/fn_tutorial_postInit.sqf
@@ -1,0 +1,60 @@
+/*
+    File: fn_tutorial_postInit.sqf
+    Author:
+    Date: 2024-11-16
+    Last Update: 2024-11-17
+    Public: No
+
+    Description:
+        PostInit for the tutorial system
+
+*/
+
+private _categories = configProperties [missionConfigFile >> "CfgHints", "isClass _x", true];
+
+{
+    private _categoryConfig = _x;
+    private _triggerableTutorials = configProperties [_categoryConfig, "isClass _x && (isClass (_x >> 'Triggers'))", true];
+    {
+        // Set up triggers for 'markerEntered' events.
+        // Can't be done in preInit, as we need `markerPos`, which is only ready in postInit (seemingly).
+        private _hintConfig = _x;
+        private _triggerConfig = _hintConfig >> "Triggers";
+        private _marker = getText (_triggerConfig >> "markerEntered");
+        if (_marker != "" && markerPos _marker isNotEqualTo [0, 0, 0]) then {
+            _marker setMarkerAlphaLocal 0;
+            private _trigger = createTrigger ["EmptyDetector", markerPos _marker, false];
+            _trigger setTriggerActivation ["ANYPLAYER", "PRESENT", true];
+            private _markerSize = markerSize _marker;
+            _trigger setTriggerArea [_markerSize # 0, _markerSize # 1, markerDir _marker, markerShape _marker isEqualTo "RECTANGLE"];
+            _trigger setTriggerStatements [
+                "this && player in thisList",
+                format ["['%1', '%2'] call vgm_c_fnc_tutorial_trigger;", configName _categoryConfig, configName _hintConfig],
+                ""
+            ];
+        };
+
+        // Create a global subscription for an event.
+        // This might need tweaking if events on other clients start triggering local tutorials.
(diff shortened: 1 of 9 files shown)


game/functions/systems/tutorial/client/fn_tutorial_postInit.sqf
```diff
@@ -0,0 +1,60 @@
...
+        private _triggerConfig = _hintConfig >> "Triggers";
+        private _marker = getText (_triggerConfig >> "markerEntered");
+        if (_marker != "" && markerPos _marker isNotEqualTo [0, 0, 0]) then {
+            _marker setMarkerAlphaLocal 0;
+            private _trigger = createTrigger ["EmptyDetector", markerPos _marker, false];
+            _trigger setTriggerActivation ["ANYPLAYER", "PRESENT", true];
+            private _markerSize = markerSize _marker;
+            _trigger setTriggerArea [_markerSize # 0, _markerSize # 1, markerDir _marker, markerShape _marker isEqualTo "RECTANGLE"];
+            _trigger setTriggerStatements [
+                "this && player in thisList",
```
veteran29: Lazy eval would be slower here most likely so we can simply omit `this`.
```suggestion
                "player in thisList",
```

game/functions/systems/tutorial/client/fn_tutorial_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+    Author:
+    Date: 2024-11-16
+    Last Update: 2024-11-17
+    Public: No
+
+    Description:
+        PreInit for the tutorial system
+
+*/
+
```
veteran29, line 14:
```suggestion

if (!hasInterface) exitWith {};
```

game/functions/systems/tutorial/client/fn_tutorial_preInit.sqf
```diff
@@ -0,0 +1,30 @@
...
+    createHashMapFromArray [
+        ["name", "TutorialAcknowledgeHint"],
+        ["displayName", "STR_VGM_TUTORIAL_UI_ACKNOWLEDGE_HINT"],
+        ["onRelease", false],
+        ["defaultKey", createHashMapFromArray [
+            ["dikCode", DIK_8]
+        ]]
+    ]
+] call para_c_fnc_keyhandler_registerAction;
+["TutorialAcknowledgeHint", para_c_fnc_ui_hints_acknowledgeHintKeypress] call para_c_fnc_keyhandler_addGeneralActionHandler;
```
veteran29, line 28: I feel like we should re-use vanilla `help` action for this.
<https://community.bistudio.com/wiki/addUserActionEventHandler>
<https://community.bistudio.com/wiki/inputAction/actions>
`help` is the action name for vanilla Adv Hints.

game/functions/systems/tutorial/client/fn_tutorial_trigger.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+    Example(s):
+        ["Tutorial", "Tutorial1"] call vgm_c_fnc_tutorial_trigger;
+ */
+
+params ["_category", "_tutorialName"];
+
+private _key = [_category, _tutorialName];
+
+if (_key in vgm_c_tutorial_seenTutorials) exitWith {};
```
veteran29, line 26: This is good enough for now but I feel like we could filter out seen tutorials during mission init and not even create a hooks for them.

veteran29 (changes requested):

game/functions/systems/tutorial/client/fn_tutorial_trigger.sqf
Spoffy: Agreed

game/functions/systems/tutorial/client/fn_tutorial_preInit.sqf
Spoffy: That seems reasonable. Do these handle being in UIs, menus, correctly?

veteran29: It seems to work only in main display and not on dialogs. Could use `actionKeysEx` to get the default value of the bind but I realized it's `H` by default which would collide with medical menu, so disregard this comment.

veteran29 (approved):

PR #204, opened 2024-12-05
Persistence - Enable saving

Spoffy:


game/functions/core/server/db/fn_db_postInit.sqf
```diff
@@ -0,0 +1,17 @@
...
+    File: fn_display_postInit.sqf
+    Author: Savage Game Design
+    Date: 2024-03-24
+    Last Update: 2024-12-05
+    Public: No
+
+    Description:
+        PostInit for database and persistence.
+ */
+
```
veteran29, line 11: I think we still need these in case we build as mission
```suggestion
if (!isServer) exitWith {};

```

veteran29 (approved):

PR #210, opened 2024-12-06
Missions - Several tracker changes

Spoffy:
1. Sets a minimum alertness of 3% to start sending trackers
2. Increases amount of alertness from shooting
3. Allows trackers to immediately start spawning once alertness crosses the threshold (no minimum time).
4. Fixes a bug causing trackers to be sent too often (up to 2 minutes early, on a 5 minute interval!!!)
5. Reworks tracking so the AI only tracks when they find your trail (not within a large radius)


game/functions/systems/mission_director/server/fn_director_processMission.sqf
```diff
@@ -41,7 +43,11 @@ private _currentTrackerDelay = linearConversion [
 private _randomExtraTime = random 120;
```
veteran29, line 43: How often processMission runs? I feel like this might be rolled fairly often, maybe it should be saved and rolled again once the first rolled time passed?

veteran29 (approved):

game/functions/systems/mission_director/server/fn_director_processMission.sqf
Spoffy: Yeah, it probably should be - this is something I spotted earlier. For now it works fine though, it's something to correct later I think :)

Spoffy: (And every 10 seconds, to answer your question)

PR #220, opened 2024-12-18
Core - Update game type and mission group

veteran29:
- title
- Will result in a save wipe on the servers as they're using file name now (vgm-latest) for the vars file


Spoffy (approved):

Spoffy:
Not merging until after our profile namespace change to make it easier to migrate

PR #221, opened 2024-12-19
Core - Switch to profile namespace for persistence

Spoffy:


game/functions/core/server/db/fn_db_clear.sqf
```diff
@@ -20,5 +20,5 @@
 */
 
 {
-	missionProfileNamespace setVariable [_x, nil];
-} forEach allVariables missionProfileNamespace;
+	profileNamespace setVariable [_x, nil];
+} forEach allVariables profileNamespace;
```
veteran29: I would filter it to clear only our vars to prevent nuking some client profile data
```suggestion
} forEach (allVariables profileNamespace select {_x find "vgm_" == 0});
```

game/functions/core/server/db/fn_db_persist.sqf
```diff
@@ -2,11 +2,11 @@
     File: fn_db_persist.sqf
     Author: Savage Game Design
     Date: 2024-12-05
-    Last Update: 2024-12-05
+    Last Update: 2024-12-19
     Public: No
 
     Description:
-        Persists the database to disk (i.e calls saveMissionProfileNamespace)
+        Persists the database to disk (i.e calls saveprofileNamespace)
```
veteran29:
```suggestion
        Persists the database to disk (i.e calls saveProfileNamespace)
```

veteran29 (changes requested):

PR #227, opened 2025-01-05
Missions - Delay shared hub enabling

veteran29:
- Delay shared hub enable until the player is teleported back into the hub
- Should prevent issues with players being damaged at the end of the mission

### DIFF
--- a/game/functions/systems/missions/client/internal/fn_missions_startDeploy.sqf
+++ b/game/functions/systems/missions/client/internal/fn_missions_startDeploy.sqf
@@ -2,7 +2,7 @@
     File: fn_missions_startDeploy.sqf
     Author: Savage Game Design
     Date: 2023-02-26
-    Last Update: 2024-11-16
+    Last Update: 2025-01-05
     Public: No
 
     Description:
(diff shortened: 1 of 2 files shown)


game/functions/systems/missions/client/internal/fn_missions_startDeploy.sqf
```diff
@@ -22,6 +22,7 @@
 
 vgm_mission_onMission = true;
 
+terminate (missionNamespace getVariable ["vgm_missions_enableHubScript", scriptNull]);
```
Spoffy, line 25: Can we move this into _disableHub? Feels like a less error-prone place for it to live than in startDeploy :)

Spoffy (changes requested):

game/functions/systems/missions/client/internal/fn_missions_startDeploy.sqf
veteran29: I don't feel like it would fit in there as the script is started in client side `endMission`. `fnc_disableHub` is a different component

PR #229, opened 2025-01-05
Leveling - Add level indicator when in Shared Hub

veteran29:
- title

image


game/functions/systems/equipment/client/fn_equipment_postInit.sqf
```diff
@@ -37,3 +37,13 @@ player call vgm_c_fnc_equipment_setDefaultLoadout;
 ["vgm_skills_respecLocal", {
     player call vgm_c_fnc_equipment_setDefaultLoadout;
 }] call para_g_fnc_event_subscribeLocal;
+
+["vgm_equipment_arsenalOpen", {
+    vgm_equipment_arsenalOpen = true;
```
Spoffy, line 42: We set this here, but I don't think we set it to 'false' again when the arsenal closes

Spoffy (changes requested):

PR #236, opened 2025-01-16
Stop, Focus - Automatically disable when strafing or shooting

Spoffy:


veteran29 (changes requested):
Don't see the code for handling strafing

veteran29 (approved):

PR #241, opened 2025-01-17
Scouting - Improved site photo detection

veteran29:


game/functions/systems/sites/server/fn_sites_postInit.sqf
```diff
@@ -19,3 +19,34 @@
...
+        private _mission = [_missionId] call vgm_s_fnc_missions_getById;
+        if !([_site, _mission] call vgm_s_fnc_sites_isInMission) exitWith {};
+
+        private _sitesData = [_missionId, "sites"] call vgm_s_fnc_missions_getSystemNetmap;
+        [_sitesData get "sites", _site get "pos", _site get "objects"] call para_s_fnc_netmap_set;
+    }]] call para_g_fnc_event_subscribeServer;
+
+    vgm_sites_siteSpawnedHandlers set [_missionId, _id];
+
+}] call para_g_fnc_event_subscribeServer;
```
Spoffy: This felt a little convoluted to me, as we're catching an event from another system, to register an event with our own system, that (effectively) tracks this object back in the original system. 

Can I suggest we look at using the `mission_zones` system for this instead? The zones already track all the sites present in that zone. We could change that to a netmap (instead of a server-side hashmap) and use that.

That has the advantages that:
- Missions are aware of zones
- Zones are aware of sites

So transitively we can already get a list of sites without additional code.

It also helps us in the future, as zones are independent of missions - i.e, zones can have data that persists between missions (it just doesn't currently).

Spoffy: `vgm_missions_zones_spawnedSites` is the variable we should probably netmap.

game/functions/systems/sites/server/fn_sites_registerMission.sqf
```diff
@@ -0,0 +1,25 @@
...
+
+    Example(s):
+        _missionId call vgm_s_fnc_sites_registerMission
+ */
+
+params ["_missionId"];
+
+private _scoutingNetmap = [_missionId, "sites"] call vgm_s_fnc_missions_createSystemNetmap;
+
+[_scoutingNetmap, "sites", [] call para_s_fnc_netmap_createNetmap] call para_s_fnc_netmap_set;
```
Spoffy: See other comment about potentially using the zones system, instead of extended sites to be mission aware.

game/functions/systems/sites_hints/server/fn_sites_hints_place.sqf
```diff
@@ -23,12 +23,13 @@
 
 params ["_mission", "_site"];
 
+// ignore sites outside of our mission zone
+// sites do not keep track of what mission they're spawned in
+if !([_site, _mission] call vgm_s_fnc_sites_isInMission) exitWith {};
```
Spoffy, line 28: Potentially switch this to using `mission_zones` system.

game/functions/systems/sites/server/fn_sites_isInMission.sqf
```diff
@@ -0,0 +1,27 @@
...
+    Example(s):
+        [_site, _mission] call vgm_s_fnc_sites_isInMission
+ */
+
+params ["_site", "_mission"];
+
+private _zone = _mission call vgm_g_fnc_missions_getZoneMarker;
+private _center = _site get "pos";
+
+_center inArea _zone // return
```
Spoffy, line 27: This feels like it could live in `missions_zones`, which already tracks all the spawned sites within a zone. 
(See big comment in `fn_sites_postinit.sqf`)

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
```diff
@@ -2,32 +2,169 @@
...
+    vgm_scouting_debug_photoLines = [];
+};
+#endif
+
+// check if position is in photo frame
+private _fnc_isInFrame = {
+    private _screenPos = worldToScreen ASLToAGL _this;
+    if (_screenPos isEqualTo []) exitWith {false};
+    _screenPos params ["_sX", "_sY"];
+    (_sX >= -0.4 && _sX <= 1.4) && (_sY >= -0.13 && _sY <= 1.13) // return
```
Spoffy, line 35: Could we add a comment here explaining this check? I'm guessing it's a "Keep in center of the screen" check, but not sure.

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
```diff
@@ -2,32 +2,169 @@
...
+            if (count _visibleObjectPoints >= 2) then {break};
+
+        } forEach (_object call _fnc_getObjectPoints);
+    } forEach _spottableObjects;
+
+    _objects // return
+};
+
+private _mission = call vgm_c_fnc_missions_getCurrentMission;
+private _extern_sites = (_mission get "system_sites" get "sites");
```
Spoffy: (See fn_sites_postInit.sqf) for full comment.

I think could be `[_mission get "targetZone"] call vgm_c_missions_zones_getSites` by changing the missions_zones to use a netmap.

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
```diff
@@ -2,32 +2,169 @@
...
+            {
+                _x params ["_beg", "_end", "_text", "_lineColor"];
+                drawLine3D [_beg, _end, _lineColor, 5];
+                drawIcon3D ["", [1,1,1,1], _end, 1, 1, 0, _text];
+            } forEach vgm_scouting_debug_photoLines;
+        }];
+        vgm_scouting_debug_drawEh = _debugEh;
+        vgm_scouting_debug_photoData = _photoData;
+    };
+    #endif
```
Spoffy:
```suggestion
#endif
```

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
```diff
@@ -2,32 +2,169 @@
...
+};
+
+private _mission = call vgm_c_fnc_missions_getCurrentMission;
+private _extern_sites = (_mission get "system_sites" get "sites");
+private _extern_player = focusOn;
+
+// gather sites that are on a screen
+private _sitesPositions = (_extern_sites call para_g_fnc_netmap_keys) apply {_x + [getTerrainHeightASL _x + 1.5]};
+private _visiblePositions = _sitesPositions select {_x call _fnc_isInFrame};
+
```
Spoffy: Would it simplify some of the following code, if you passed site hashmaps around rather than positions? E.g:

```suggestion
private _sites = _extern_sites call para_g_fnc_netmap_values;
private _visibleSites = _sites select {((_x get "pos") + [getTerrainHeightASL _x + 1.5]) call _fnc_isInFrame};
```

_fnc_getForegroundObjects wouldn't need the hashmap lookup then to get the site, and _photoData could use `_site get "id"` as an index.

Spoffy (changes requested):

game/functions/systems/missions_gameplay/client/scouting/fn_missions_gameplay_scouting_onPhoto.sqf
veteran29: The comment above the function is `// check if position is in photo frame` so that's basically what that is... 😅

game/functions/systems/sites/server/fn_sites_postInit.sqf
veteran29: Refactored to use a netmap from missions_zones

game/functions/systems/sites_hints/server/fn_sites_hints_place.sqf
veteran29: The issue with this is that hints are placed based on `vgm_sites_siteSpawned` event while mission zones uses returns from site spawn function. So `missions_zones` is not yet aware about sites at this point.

game/functions/systems/sites/server/fn_sites_isInMission.sqf
veteran29: See the response in [fn_sites_hints_place.sqf](https://github.com/Savage-Game-Design/Across-the-Fence/pull/241/files/5d3b13ea145377bfedb1b13f5d9d82dd2a89c582#diff-ccd8ee8f5c30024f24b92f12a84c9cca73e4927ea409c4b4a8db8b66af6ee046)

Spoffy (approved):

PR #255, opened 2025-02-03
Focus - Add enable buffer time, refactor

veteran29:
- 0.4s buffer time while checking for "canFocus"
- Addition of `waitUntilAndExecute` function with same API as [CBA](https://cbateam.github.io/CBA_A3/docs/files/common/fnc_waitUntilAndExecute-sqf.html) one
- Refactor of Focus loop to utilise `waitUntilAndExecute`
- "Error" hint/sound is not played anymore when focus is disabled due to movement, only when unable to enable due to movement, IMO it's bit better as hearing the error sound every time you break focus by moving is annoying

### DIFF
--- a/game/functions/core/global/fn_waitUntilAndExecute.sqf
+++ b/game/functions/core/global/fn_waitUntilAndExecute.sqf
@@ -1,0 +1,61 @@
+/*
+    File: fn_waitUntilAndExecute.sqf
+    Author: Savage Game Design
+    Date: 2025-02-02
+    Last Update: 2025-02-03
+    Public: No
+
+    Description:
+        No description added yet.
+
+    Parameter(s):
+        N/A
+
+    Returns:
+        Something [BOOL]
+
+    Example(s):
+        [parameter] call vgm_X_fnc_component_myFunction
+ */
+
+#define REMOVE_ITEM vgm_core_waitUntilAndExecute_array deleteAt _forEachIndex;
+
+params [
+    ["_condition", {true}, [{}]],
+    ["_code", {}, [{}]],
+    ["_args", []],
+    ["_timeoutDelay", nil, [0]],
+    ["_timeoutCode", {}, [{}]]
+];
+
+if (vgm_core_waitUntilAndExecute_eh == -1) then {
+    private _eh = addMissionEventHandler ["EachFrame", {
+        {
+            _x params ["_condition", "_code", "_args", "_timeoutTime", "_timeoutCode"];
+            // if _timeoutDelay is nil this will be not be executed due to how SQF operators work
+            if (time > _timeoutTime) then {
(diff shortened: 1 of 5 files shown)


game/functions/core/global/fn_waitUntilAndExecute.sqf
```diff
@@ -0,0 +1,73 @@
...
+            if (_args call _condition) then {
+                _args call _code;
+
+                REMOVE_ITEM;
+                continue;
+            };
+        } forEachReversed vgm_core_waitUntilAndExecute_array;
+
+        if (vgm_core_waitUntilAndExecute_array isEqualTo []) then {
+            // we could consider having the EachFrame run all the time to prevent constant increments of the EH ids
```
Spoffy: We could just run it all the time - I'd be happy if you want to simplify that. Most of the time it would just be doing:
```
{} forEachReversed []
```

game/functions/systems/skill_investigate/client/fn_skill_investigate_setFocusMode.sqf
```diff
@@ -23,35 +23,39 @@
...
+                vgm_c_skill_investigate_isFocusing = false;
+                [false] call vgm_c_fnc_skill_investigate_setDesaturation;
+                [false] call vgm_c_fnc_skill_investigate_setListenMode;
+            },
+            _this
+        ] call vgm_g_fnc_waitUntilAndExecute;
+    },
+    nil,
+    0.4,
+    {
```
Spoffy:
```suggestion
[
    // Condition to wait for
    {call vgm_c_fnc_skill_investigate_canFocus},
    // Code to execute
    {
        [true] call vgm_c_fnc_skill_investigate_setDesaturation;
        [true] call vgm_c_fnc_skill_investigate_setListenMode;

        // check if player can remain focused
        [
            {
                !vgm_c_skill_investigate_isFocusing
                || !(call vgm_c_fnc_skill_investigate_canFocus)
            },
            {
                vgm_c_skill_investigate_isFocusing = false;
                [false] call vgm_c_fnc_skill_investigate_setDesaturation;
                [false] call vgm_c_fnc_skill_investigate_setListenMode;
            },
            _this
        ] call vgm_g_fnc_waitUntilAndExecute;
    },
    // Arguments
    nil,
    // Timeout
    0.4,
    // Code executed on timeout
    {
```
Optional suggestion, but these comments might help with readability for someone seeing this for the first time. 🙂

Spoffy (approved):
Approved - this is a neat solution, didn't know about that CBA method. Nice use for `forEachReversed` for the safe removal while iterating too. 

Left a suggestion to help with understandability for someone looking at this fresh, but it's 100% optional.

PR #256, opened 2025-02-03
Focus - Use player animation state for condition check

veteran29:


game/functions/systems/skill_investigate/client/fn_skill_investigate_canFocus.sqf
```diff
@@ -18,4 +18,10 @@
...
  */
 
-vectorMagnitudeSqr velocity player < 1
+// bipod anims have varying names so simply check if it's deployed
+isWeaponDeployed player
+|| {
+    animationState player select [9, 3] isEqualTo "stp"
+    // checking animation "step" allows us to catch "odd" ones like rolling while prone
+    && vectorMagnitude (player getUnitMovesInfo 4) < 1 // return
+} // return
```
Spoffy, line 27: Can we tweak this so it allows "Stop, Focus" when you're on a static gun? E.g the AA guns? It's not strictly necessary, but would be a nice to have, as I've seen people using the AA guns from time to time. 

The check that comes to mind to me is:
```
|| vehicle player != player && vectorMagnitudeSqr velocity player == 0 && !(isEngineOn vehicle player)
```
Since the vehicles use custom animations we can't check those. But this checks at least that they're still, and the engine isn't on (since the engine would be loud). 

Open to better ideas though :)

Spoffy: If you want to merge this, and do it in a separate PR, I'm also fine with that, since it's basically an extension of this.

Spoffy (changes requested):

game/functions/systems/skill_investigate/client/fn_skill_investigate_canFocus.sqf
veteran29: Feels like something for a separate PR

PR #257, opened 2025-02-06
Medical - Fix unconscious state sometimes not recognized on the server

veteran29:


game/functions/systems/medical/global/fn_medical_isUnconscious.sqf
```diff
@@ -0,0 +1,23 @@
...
+    Returns:
+        Unconscious state [BOOL]
+
+    Example(s):
+        player call vgm_g_fnc_medical_isUnconscious
+ */
+
+params ["_unit"];
+
+_unit getVariable "vgm_g_medical_isUnconscious" // return
```
Spoffy, line 24:
```suggestion
// lifeState can't be used due to an engine bug, which results in the server's `lifeState` being rarely being out of sync with the client's, if they take damage just before being set unconscious. This results in  `INCAPACITATED` on the client, but `INJURED` instead of `INCAPACITATED` on the server. This only lasts for a few frames, but is enough to cause issues.
_unit getVariable "vgm_g_medical_isUnconscious" // return
```

game/functions/systems/medical/client/fn_medical_setUnconscious.sqf
```diff
@@ -21,18 +21,19 @@
 
 params ["_unit", ["_state", true]];
 
-private _previousLifeState = lifeState _unit;
-
 _unit setUnconscious _state;
 _unit setCaptive _state;
 
-if (lifeState _unit != _previousLifeState) then {
+private _previousState = _unit getVariable "vgm_g_medical_isUnconscious";
```
Spoffy, line 28:
```suggestion
// Check "vgm_c_fnc_medical_isUnconscious" for why `lifeState` can't be used
private _previousState = _unit getVariable "vgm_g_medical_isUnconscious";
```

game/functions/systems/medical/global/fn_medical_isUnconscious.sqf
Spoffy: Just in case someone from the community comes along later and goes "Oh, but why didn't they just use lifeState? Let me refactor it"

Spoffy (changes requested):
Given how many `alive _x && !(_x call vgm_g_fnc_medical_isUnconscious)` checks there are, I almost wonder if we should have a `vgm_g_fnc_medical_isReady` function or similar. 

Not worth adding to this PR anyway, but if we have anything else to check in these loops in the future, that might be a nice bit of cleaning to do.

Spoffy:
Happy for this to be merged, once the comments are addressed. Nothing in here I consider a blocker :)

PR #266, opened 2025-03-01
Adds `posIndex_getItems`

Spoffy:
Provides a way to get all the items from a position index, without any `nil` entries that are caused by items being deleted from the index.

Not currently needed - but I did all the thinking on this, so thought I'd add it just in case.

### DIFF
--- a/game/functions/core/global/position_index/fn_posindex_add.sqf
+++ b/game/functions/core/global/position_index/fn_posindex_add.sqf
@@ -2,7 +2,7 @@
     File: fn_posindex_add.sqf
     Author: Savage Game Design
     Date: 2025-01-09
-    Last Update: 2025-01-24
+    Last Update: 2025-03-01
     Public: Yes
 
     Description:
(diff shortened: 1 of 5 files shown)


veteran29 (approved):
LGTM

I wonder if it's actually worth having another data structure in memory. Since the introduction of SimpleVM optimizations something like `_index get "items" select {!isNil "_x"}` should be pretty fast.

Spoffy:
> LGTM
> 
> I wonder if it's actually worth having another data structure in memory. Since the introduction of SimpleVM optimizations something like `_index get "items" select {!isNil "_x"}` should be pretty fast.

**Copied from Discord so we've got a record on this testing on the PR**

Just tested the SimpleVM question you had @veteran29, on whether this:
`_index get "items" select {!isNil "_x"}` 
is fast compared to 
`values _someHashmap`.

These are test results:
- A HashMap with 500 entries **(0.0036ms)**
- An array with 1000 entries, half of which are nil. Tested with `y select {!isNil "_x"};`.  **(0.3091ms)**
- An array with 500 entries, none of which are nil. **(0.1500ms)**

So the hashmap clocks in at around 100x faster in this instance.

With 5000 entries, the gap narrows to about 60x faster.
With 50 entries, it stays around 60x faster for the array with nils, and 30x faster for the array without nils.

PR #270, opened 2025-03-06
Makes trackers continue after the trail ends, and search the area.

Spoffy:
Extends tracker behaviour, so that they run a random distance past the end of the trail, then search the area around them.


game/functions/systems/behaviour_trees/nodes/global/fn_btree_action_tracking_findNearbyTracks.sqf
```diff
@@ -0,0 +1,62 @@
...
+    File: fn_btree_action_tracking_findNearbyTracks.sqf
+    Author: Savage Game Design
+    Date: 2024-02-02
+    Last Update: 2025-03-06
+    Public: Yes
+
+    Description:
+        Action node.
+
+        Checks for nearby tracks from the tracking system, and stores them in the backboard.
```
veteran29:
```suggestion
        Checks for nearby tracks from the tracking system, and stores them in the blackboard.
```

game/functions/systems/behaviour_trees/nodes/global/fn_btree_action_tracking_followInferredTrail.sqf
```diff
@@ -0,0 +1,75 @@
...
+    // Vector pointing from the last track to the track that came before it.
+    private _lastTrackPrevTrackVector = _lastTrack get "prevTrackVector";
+
+    if (isNil "_lastTrackPrevTrackVector") exitWith {
+        [ RESULT_FAILED ]
+    };
+
+    private _nodeParams = [_node] call vgm_g_fnc_btree_getNodeParams;
+    // Cache these values from nodeParams, so we can use sensible defaults. _state is also cheaper to access regularly.
+    private _distance = _nodeParams getOrDefaultCall ["distance", { random [0, 50, 100] }];
```
veteran29, line 43: The value needs to be set if we want to actually cache it.
```suggestion
    private _distance = _nodeParams getOrDefaultCall ["distance", { random [0, 50, 100] }, true];
```

game/functions/systems/behaviour_trees/nodes/global/fn_btree_decorator_tracking_hasFollowableNearbyTracks.sqf
```diff
@@ -44,7 +44,14 @@ _decorator set ["condition", {
 
     if (_nearbyTracks isEqualTo []) exitWith { false };
 
-    _extern_blackboard set ["tracking_currentTrack", _nearbyTracks # -1 # 0];
+    private _nextTrack = _nearbyTracks # -1 # 0;
+
+	// Only accept newer tracks as interesting, to prevent getting stuck going in circles or infinitely tracking.
```
veteran29:
```suggestion
    // Only accept newer tracks as interesting, to prevent getting stuck going in circles or infinitely tracking.
```

game/functions/core/global/fn_manWouldCollideAtPosition.sqf
```diff
@@ -0,0 +1,30 @@
+/*
```
veteran29, line 1: Is this used anywhere?

veteran29 (changes requested):

game/functions/systems/behaviour_trees/nodes/global/fn_btree_action_tracking_followInferredTrail.sqf
Spoffy: Setting these values doesn't work - every call to getNodeParams creates a new hashmap (as some of the params are functions).

Normally this gets cached in `_state` or `_extern_blackboard` (depending on what's needed), but we don't actually need `_distance` cached here - we cache `_destination` instead which is all we need :slightly_smiling_face: 

Also, this particular function (onEnter) only runs once per node execution - and because of the nature of the node, it's fine to not worry about caching the random value here I think.

game/functions/core/global/fn_manWouldCollideAtPosition.sqf
Spoffy: Nope. Built and tested this thinking I'd need it, but right now I don't - used `findEmptyPosition` instead.

I'm probably going to use this in `patrolArea` though to know when to skip patrol points.

veteran29 (approved):

PR #278, opened 2025-04-09
Core - Enable "AIThinkOnlyLocal" mission option

veteran29:
Without 2.20 it will be simply ignored


Spoffy:
I think this might break the stealth system. Need to investigate.

Spoffy:
Probably okay actually - we use `targets` in the AI framework, but that's always local to the AI. Stealth system just uses visibility checks, as far as I can see.

PR #297, opened 2025-07-26
Builds: Launched processes interruptible from shell.

dijksterhuis:
Adds `--interrupt` and `--polling` options to `launch` command group.

When `--interrupt` is provided, CLI will track the state of launched processes and will interrupt them when:
* SIGINT received (Ctrl +C in shell)
* at least one of the processes has exited (quitting the sever manually will quit the client)
* some error occurred when managing the process (unlikely, but there as an edge case)

`--polling` can be used to set the polling duration of checking process state (seconds)

### DIFF
--- a/build/vgm/arma.py
+++ b/build/vgm/arma.py
@@ -48,7 +48,7 @@ def launch(arma_exe_path, mods=[], args=[], connect=False, editor_mission_path: 
     print(f"Launching Arma: {list(map(str, command))}")
 
     # start arma
-    subprocess.Popen(command)
+    return subprocess.Popen(command)
 
 def launch_server(mission_path: Path, arma_server_exe: Path, config: Path, servermod_path=None, mods=[], profile="vgm_server"):
     try_symlink_arma_server_mission_dir(mission_path, arma_server_exe)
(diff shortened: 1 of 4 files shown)


build/vgm/arma.py
```diff
@@ -48,24 +48,24 @@ def launch(arma_exe_path, mods=[], args=[], connect=False, editor_mission_path:
...
+    return subprocess.Popen(command)
 
 def launch_server(mission_path: Path, arma_server_exe: Path, config: Path, servermod_path=None, mods=[], profile="vgm_server"):
     try_symlink_arma_server_mission_dir(mission_path, arma_server_exe)
 
     server_config_path = setup_temporary_arma_server_config(config, mission_path.name)
 
     if servermod_path:
-        hemtt.launch(
+        return hemtt.launch(
```
dijksterhuis, line 59: uses `subprocess.run` (needs a `Popen`)

dijksterhuis:
@Spoffy i wouldn't mind also adding a `--tail` option here as well (conditionally `tail -f` the server/client log file in terminal while processes are running).

will need an extra optional config option (`arma3_logs_dirpath`) but that's about it.

can either add in here, or a separate PR once this is merged.

build/vgm/processes.py
```diff
@@ -0,0 +1,86 @@
+"""
+Process management utility functions.
+"""
+
+from subprocess import Popen
+from time import sleep
+from typing import List
+
+
+def check_terminate_process(process: Popen, timeout: int = 15):
```
Spoffy: Could probably just call this `terminate_process`, not sure `check` helps with understanding what this func does :slightly_smiling_face:

Spoffy (approved):

Spoffy:
Separate PR for any tail behaviour please :)

Although that also feels like it can wait for a bit :)

build/vgm/processes.py
dijksterhuis: done -- also moved `check_alive` higher than `terminate_processes` (it appears first in the `process_handler` function)

dijksterhuis:
> Separate PR for any tail behaviour please :)
> 
> Although that also feels like it can wait for a bit :)

wilco -- will build off this branch so will be a dependent PR

PR #299, opened 2025-07-28
Build: 'Fix' windows path handling for server config argument.

dijksterhuis:
Caveat: This may be a "@dijksterhuis is using a weird terminal environment" problem.

----

When trying to run `python3 ./build/run.py launch dev --overwrite` or `python3 ./build/run.py launch server` in a MSYS2 (MGWIN64) shell, the launched arma server refuses to load the generated server configuration file. Which then means battleye is enabled and all sorts.

`pathlib.Path` paths in an MSYS2 environment are actually `pathlib.WindowsPath`s under the hood. But when the path is represented in an f-string, python outputs forward slashes.

For the path to the arma executables etc, this is fine. however, Arma CLI args require proper non-posix windows path structure a la `DRIVE:\DIR\DIR\DIR\FILENAME.EXT` ..... i.e. only backslashes.

### DIFF
--- a/build/vgm/arma.py
+++ b/build/vgm/arma.py
@@ -1,8 +1,28 @@
 import subprocess
 import sgd.file_utils
 from pathlib import Path
+from sys import platform
 
 from . import hemtt
+
+
+def maybe_fix_win_path(path: Path):
+    """
+    Fix path borkage for Arma CLI args on Windows
+
+    Notes:
+        Arma requires file path CLI arguments look exactly like this:
+        `DRIVE:\DIR\DIR\DIR\FILENAME.EXT`
+
+        `pathlib.Path` reprs to `DRIVE:/DIR/DIR/FIR/FILENAME.EXT`, which
+        Arma's CLI won't recognise as a valid Windows path. Escaped paths
+        a la `DRIVE:\\DIR\\DIR\\DIR\\FILENAME.EXT` also do not work.
+
+        If you're tweaking the arma_server.hpp configuration file and the
+        builder CLI is not updating your server with new configuration options,
+        this is probably your exact problem.
+    """
+    return f"{path}".replace("/", "\\") if platform in ["win32", "cygwin"] else path
 
 def find_mpmissions(search_start: Path) -> Path | None:
     def is_mpmissions(path: Path) -> bool:
@@ -53,7 +73,9 @@ def launch(arma_exe_path, mods=[], args=[], connect=False, editor_mission_path: 
 def launch_server(mission_path: Path, arma_server_exe: Path, config: Path, servermod_path=None, mods=[], profile="vgm_server"):
     try_symlink_arma_server_mission_dir(mission_path, arma_server_exe)
 
-    server_config_path = setup_temporary_arma_server_config(config, mission_path.name)
+    server_config_path = maybe_fix_win_path(
+        setup_temporary_arma_server_config(config, mission_path.name)
+    )
 
     if servermod_path:
         hemtt.launch(
--- a/build/vgm/arma.py
+++ b/build/vgm/arma.py
@@ -7,7 +7,7 @@ from . import hemtt
 
 
 def maybe_fix_win_path(path: Path):
-    """
+    r"""
     Fix path borkage for Arma CLI args on Windows
 
     Notes:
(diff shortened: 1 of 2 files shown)


dijksterhuis:
@Spoffy when reviewing please could you check it doesn't bork your local build?

build/vgm/arma.py
```diff
@@ -1,9 +1,28 @@
...
+        `DRIVE:\DIR\DIR\DIR\FILENAME.EXT`
+
+        `pathlib.Path` reprs to `DRIVE:/DIR/DIR/FIR/FILENAME.EXT`, which
+        Arma's CLI won't recognise as a valid Windows path.
+
+        If you're tweaking the arma_server.hpp configuration file and the
+        builder CLI is not updating your server with new configuration options,
+        this is probably your exact problem.
+    """
+    return f"{path}".replace("/", "\\") if platform in ["win32", "cygwin"] else path
```
Spoffy: Instead of mangling the path manually, could you try using `PureWindowsPath` instead? It might behave correctly despite MinGW. I think you'll still want the `win32`/`cygwin` check in?

Could we also rename this to `force_windows_path_on_windows` or similar?

Spoffy (changes requested):
Definitely feels like a "DJ is using a wonky terminal" problem... possibly Python reading the MinGW environment as a Unix one instead of Windows, and outputting for Unix? Wouldn't surprise me if it's got an explicit edge case built in.

build/vgm/arma.py
dijksterhuis: > Instead of mangling the path manually, could you try using `PureWindowsPath` instead? It might behave correctly despite MinGW.

`PureWindowsPath` doesn't work unfortunately, python will still repr the path as posix when f-string'd for process arguments. 
```python3
from pathlib import PureWindowsPath
p = PureWindowsPath(r"C:\Users\dijksterhuis\Documents\MIKE-FORCE\SGD\atf\build\arma_server.tmp")
f"{p}"
# 'C:/Users/dijksterhuis/Documents/MIKE-FORCE/SGD/atf/build/arma_server.tmp'
f"{p}".replace("/", "\\")
# 'C:\\Users\\dijksterhuis\\Documents\\MIKE-FORCE\\SGD\\atf\\build\\arma_server.tmp'
```

> I think you'll still want the win32/cygwin check in?

I wasn't sure about `cygwin` when I added it, but that's also posix path based so yep will have the same issue.

Windows dev is fun...

> Could we also rename this to `force_windows_path_on_windows` or similar?

sure, i'd prefer a verb in there if we can like create/make/etc (implies a return value)

- `force_create_windows_path_on_windows` ?
- `create_explicit_windows_path_on_windows` ?

dijksterhuis:
> Definitely feels like a "DJ is using a wonky terminal" problem... 

eh, i'd argue it's more arma only supporting one possible path variant in windows -- windows path delimiters are either `\` or `/`
https://en.wikipedia.org/wiki/Path_(computing)#Path_syntax 

**_Plus_**, my *artisanal* preferences :D (when in doubt, blame arma, but point taken, i still hate powershell/cmd)

> possibly Python reading the MinGW environment as a Unix one instead of Windows, and outputting for Unix? Wouldn't surprise me if it's got an explicit edge case built in.

technically, no -- the `C:/path/to/some/file.txt` path is a valid windows path, but arma doesn't support it. case in point -- `X:/path/to/arma3.exe` will execute fine in CMD / powershell.

examples in here: https://www.msys2.org/docs/filesystem-paths/

always felt like the fact the executable ran fine with `/` delimiters was odd (i have learned something today about path syntax)

dijksterhuis:
for documentation

GCC python (MSYS2) vs MSC python (CMD)
image

MSC python run from MSYS2
image

Spoffy (approved):

dijksterhuis:
hemtt might need this too. 

hemtt launch refused to load server mod. i bet its the same deal (not tested, AFK)

Spoffy:
Double check the path we're passing to hemtt is correct! But it would be weird if that was the case, since I'm guessing you're running the Windows HEMTT binary?

dijksterhuis:
i had to build from source because of course i had to build it from source 😂

powershell is starting to look like something i'm going to have to drop my ideological hatred for very quickly 😂

dijksterhuis:
looks like could be a different issue. getting the error on CMD with winget's hemtt with all custom development mods disabled in the config (zeus utils stuff etc). the `server_mod_path` path looks `\` delimited valid (and it contains the built mod files).
```bat
$ python3 ./build/run.py launch server --mod
DEBUG_SERVERMODPATH C:\Users\dijksterhuis\Documents\MIKE-FORCE\SGD\atf\output\mods\@vgm_server
Launching HEMTT: ['hemtt', 'launch', '--', '-name=vgm_server', '-config=C:\\Users\\dijksterhuis\\Documents\\MIKE-FORCE\\SGD\\atf\\build\\arma_server.tmp']

# does build

error[BCLE2]: Arma 3 workshop mod `1649778181` not found.
 = help: HEMTT does not subscribe to mods, you must subscribe in Steam and allow it to download.
         Workshop link: https://steamcommunity.com/sharedfiles/filedetails/?id=1649778181
```

will merge PR then debug this separately.

PR #309, opened 2025-08-21
Tutorials: Desperate Escape.

dijksterhuis:
Needs actual text, and i'm not 100% sure how the toml tutorial builds work

### DIFF
--- a/game/configs/mission/cfg_field_manual.hpp
+++ b/game/configs/mission/cfg_field_manual.hpp
@@ -195,6 +195,7 @@ class vgm_missions {
     class desperate_escape {
         displayName = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE";
         description = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE_DESC";
+        tip = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE_TIP";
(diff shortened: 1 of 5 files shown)


game/configs/mission/cfg_field_manual.hpp
```diff
@@ -192,4 +192,16 @@ class vgm_missions {
...
+    class desperate_escape {
+        displayName = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE";
+        description = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE_DESC";
+        image = "assets\atf_logo.paa";
+        class Hints {
+            class desperate_escape {
+                description = "$STR_VGM_FIELD_MANUAL_DESPERATE_ESCAPE_HINT_DESPERATE_ESCAPE_DESCRIPTION";
+            };
+        };
+        logicalOrder = 6;
```
dijksterhuis, line 205: no idea what this is for, i assume it's supposed to increment as per other entries

Spoffy: Specifies the order these are put in in the field manual :)

Spoffy (approved):
I updated the text on here. We also don't modify cfg_field_manual directly, it's in `index.toml`. I forgot that too, so maybe it needs a big fat comment in there warning about it.

PR #311, opened 2025-08-23
Respawn: Simplify search and fix players spawning near enemies.

dijksterhuis:
change in search discussed with @Spoffy, thanks to @veteran29 for spotting the 2D and AGLtoASL issue.

### DIFF
--- a/game/functions/systems/respawn/global/fn_respawn_findSafeSpawnTransformNearTeam.sqf
+++ b/game/functions/systems/respawn/global/fn_respawn_findSafeSpawnTransformNearTeam.sqf
@@ -65,24 +65,35 @@ private _groupPositionAGL = [
     ]
 ] call para_g_fnc_get_group_majority_position;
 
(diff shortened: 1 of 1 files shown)


dijksterhuis:
in draft until i've tested it

game/functions/systems/respawn/global/fn_respawn_findSafeSpawnTransformNearTeam.sqf
```diff
@@ -65,24 +65,35 @@ private _groupPositionAGL = [
     ]
 ] call para_g_fnc_get_group_majority_position;
 
-if (_groupPositionAGL isEqualTo [0, 0, 0]) exitWith { // couldn't determine group position (maybe they're all dead?), use fallback spawn transform
```
dijksterhuis, line 68: i noticed that single players wouldn't get a search previously, so fixed that alongside the others.

Spoffy (approved):
Approved once tested :)

dijksterhuis:
@Spoffy looks like it's working a-ok. no script debug errors at the very least (and i'm not spawning near enemies).

before escape



after escape

PR #312, opened 2025-08-23
Extraction: Immersive helo extraction to TOC on mission end.

dijksterhuis:
Fix for player freefalling on mission end which doesn't need a dramatic/cinematic helo ride -- teleport the helo then move the players.


dijksterhuis:
ok,. no, doesn't work reliably. needs fade in/out and a sleep. so basically one half of the other PR.

dijksterhuis:
note -- may want this at a later date for "cooler" extractions so re-opening

dijksterhuis:
~~i would love to put this back into draft ... but github menus are confusing and non obvious :/~~ found the button

PR #341, opened 2025-09-10
Lobby: Reconfigure slots to represent 6 different teams.

dijksterhuis:
Change types of player slots to create pseudo team allocations. A bit more varied and closer to reality, rather than everyone being an RTO.

Also re-organize the player objects to make updating slots easier in the future (server owners may want to add more mission teams -- _just copy and paste the last row_).

### editor re-org

this took quite a while ... all the "teams" are organised from back to front by their mission object ID (which translates to the order slots are shown in)

6x teams - 6x rows.

20813F~1

20F684~1

### TODO

- [x] deal with the phantom "RTO" slot -- was in there before.

### DIFF
--- a/game/maps/cam_lao_nam/mission.sqm
+++ b/game/maps/cam_lao_nam/mission.sqm
@@ -310,7 +310,7 @@ class Mission
 	};
 	class Entities
 	{
-		items=60;
+		items=59;
 		class Item0
 		{
 			dataType="Logic";
@@ -370,7 +370,7 @@ class Mission
 							dataType="Object";
 							class PositionInfo
 							{
-								position[]={20097.52,4.996995,6632.1831};
+								position[]={20093.584,5.0536008,6636.4736};
 								angles[]={0,2.285785,0};
 							};
 							side="West";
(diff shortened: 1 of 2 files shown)


game/maps/cam_lao_nam/mission.sqm
```diff
@@ -2321,72 +2453,6 @@ class Mission
...
-						nAttributes=2;
-					};
-				};
-			};
-			class Attributes
-			{
-			};
-			id=1371;
-			atlOffset=-4.7683716e-07;
-		};
```
dijksterhuis, line 2388: this was not attached to the group and would create an "ungrouped" slot at the bottom of the lobby

game/mission/description.ext
```diff
@@ -10,7 +10,7 @@ class Header
 {
     gameType = "Coop";
     minPlayers = 1;
-    maxPlayers = 24;
+    maxPlayers = 36;
```
dijksterhuis, line 13: @Spoffy important to note -- don't know if there's been comms gone out about a specific player limit (i seem to remember 6 teams of 6?)

dijksterhuis:
done by @Spoffy separately

PR #348, opened 2025-09-25
Equipment: Rifleman: Add L1A1 shorty magazines to arsenal whitelist.

dijksterhuis:


Spoffy:
This is all done via a central doc, I need to document the process really.

Spoffy:
Fixed in #357 - I replaced the whole L1A1 shorty with an L2A1 to remove the nickel steel dependency.

PR #349, opened 2025-09-25
Equipment: Friend Or Foe: Add pith helmets to arsenal whitelist.

dijksterhuis:
20E27C~1


Spoffy:
Needs to be done in the Grist doc then copied in.

Also - this should only be the Recon pith helmet, I forget which that is (Rob's idea!)

Spoffy:
Done in #357 - also needed a small code change to make the SOG Pith helmet work for "Friend or Foe"

PR #351, opened 2025-09-25
Equipment: Scout: Fix M3A1 suppressor typo.

dijksterhuis:
#343 will fix, but this makes everything lowercase and is ultimately the pedantic solution.


Spoffy:
Would be overwritten by the main document next time this is updated. Will update the class name in the data dump instead to fix this!

Spoffy:
Fixed by Veteran as you mentioned, but I also corrected this in #357 for your sanity's sake :D

PR #360, opened 2025-10-18
Skills - Fix learn conditions locality

veteran29:
- Fix locality of skills unlock conditions
- rename `conditionUnlock` to `conditionsUnlockGlobal` to clearly indicate this one is executed on server too, all other conditions are client side

### DIFF
--- a/game/functions/systems/skills/client/network/fn_skills_requestSkillRespec.sqf
+++ b/game/functions/systems/skills/client/network/fn_skills_requestSkillRespec.sqf
@@ -2,7 +2,7 @@
     File: fn_skills_requestSkillLearn.sqf
     Author: Savage Game Deisgn
     Date: 2023-02-26
-    Last Update: 2024-07-04
+    Last Update: 2025-10-18
     Public: No
 
     Description:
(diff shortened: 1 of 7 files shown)


game/functions/systems/skills/client/network/fn_skills_requestSkillRespec.sqf
```diff
@@ -20,6 +20,6 @@
 
 params [["_displayParent", nil]];
 
-["Waiting for server...", "Please wait", false, false, _displayParent] spawn BIS_fnc_guiMessage;
+isNil {["Waiting for server...", "Please wait", false, false, _displayParent] spawn BIS_fnc_guiMessage};
```
Spoffy, line 23: Why does this one need to be unscheduled? Don't mind the change, just curious.

Spoffy (approved):

game/functions/systems/skills/client/network/fn_skills_requestSkillRespec.sqf
veteran29: When `fn_skills_requestSkillRespec` is called from scheduled this line will warn/error about `_displayParent` being `nil` so this basically silences the error dialog.

PR #370, opened 2025-11-14
Adds 'codeApplyGroup' and 'codeUnapplyGroup' to skills

Spoffy:


game/functions/systems/skills/client/fn_skills_unapplyPlayersGroupSkills.sqf
```diff
@@ -0,0 +1,31 @@
...
+params ["_owningPlayerId"];
+
+private _skills = values (vgm_c_skills_appliedGroupSkills getOrDefault [_owningPlayerId, createHashMap]);
+
+{
+    player call (_x get "codeUnapplyGroup");
+} forEach _skills;
+
+vgm_c_skills_appliedGroupSkills deleteAt _owningPlayerId;
+vgm_c_skills_applyOnRespawnGroup deleteAt _owningPlayerId;
```
veteran29, line 31: Feels like this should call `fn_skills_unapplyGroupSkill`

Spoffy: Don't mind - I can see the maintainability benefit of it, and the performance impact should be negligible given this doesn't happen often.

PR #372, opened 2025-11-19
Adds optional durations to status effects

Spoffy:
Allows status effect reasons to have a duration set, and be automatically removed after a set period of time.

### DIFF
--- a/game/functions/core/client/statusEffect/fn_statusEffect_postInit.sqf
+++ b/game/functions/core/client/statusEffect/fn_statusEffect_postInit.sqf
@@ -2,7 +2,7 @@
     File: fn_statusEffect_preInit.sqf
     Author: Savage Game Design
     Date: 2024-12-07
-    Last Update: 2024-12-14
+    Last Update: 2025-11-19
     Public: No
 
     Description:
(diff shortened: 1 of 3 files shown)


game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
```diff
@@ -47,10 +59,21 @@ if (isNil "_effectsMap") then {
     }]
 };
 
-format ["Adding status effect reason: %1 | %2", _effect, _reason] call vgm_g_fnc_logDebug;
+format ["Adding status effect reason: %1 | %2 | %3", _effect, _reason, _duration] call vgm_g_fnc_logDebug;
 
 private _reasonList = _effectsMap getOrDefault [_effect, [], true];
 
+private _effectsReasonEndTimes = _effectsEndTimes getOrDefault [_effect, createHashMap, true];
```
veteran29: `_effectReasonsEndTimes` seems like a more fitting name as this hashmap holds multiple reasons for a single status effect.

```suggestion
private _effectReasonsEndTimes = _effectsEndTimes getOrDefault [_effect, createHashMap, true];
```

game/functions/core/client/statusEffect/fn_statusEffect_postInit.sqf
```diff
@@ -17,3 +17,26 @@ private _fnc_setCoreStatusEffects = {
...
+
+vgm_c_statusEffect_eachFrameEH = addMissionEventHandler ["EachFrame", {
+    // Queue of end times for status effect reasons, sorted by end time.
+    private _endTimesQueue = player getVariable ["vgm_c_statusEffect_endTimesQueue", []];
+    if (_endTimesQueue isEqualTo []) exitWith {};
+
+    // First item in the queue will always be the soonest.
+    _endTimesQueue # 0 params ["_nextEndTime", "_effectDetails"];
+
+    if (serverTime < _nextEndTime) exitWith {};
```
veteran29: I would use `time` which is tracked since mission started and affected by accTime in SP (does not matter for us though).
You usually want to avoid server time as it is more prone to floating point errors.

```suggestion
    if (time < _nextEndTime) exitWith {};
```

game/functions/core/client/statusEffect/fn_statusEffect_set.sqf
```diff
@@ -47,10 +59,21 @@ if (isNil "_effectsMap") then {
...
+if (_duration > 0) then {
+    private _endTime = serverTime + _duration;
```
veteran29: I would use `time` which is tracked since mission started and affected by accTime in SP (does not matter for us though).
You usually want to avoid server time as it is more prone to floating point errors.

```suggestion
    private _endTime = time + _duration;
```

veteran29 (changes requested):

